PR #79, opened 2014-08-14
Ocaml support

xcthulhu:


spencerwi:
What's holding this back from being available at Codewars.com? It's been 2 years and 4 days since this was merged in, but still no OCaml support there.

jhoffner:
`solutionOnly` mode is supported but not `testIntegration` mode. Want to help? The biggest thing involved is setting up the test framework and getting it to output the correct format.

spencerwi:
Ah, that's a distinction I wasn't aware of. Guess it pays to actually read the README. I may spend some time seeing if I can get a test framework up and going.

PR #122, opened 2014-09-23
Clojure Fixed

xcthulhu:
This ought to work now

### DIFF
--- a/frameworks/swift/Podfile
+++ b/frameworks/swift/Podfile
@@ -1,0 +1,1 @@
+pod 'Sleipnir'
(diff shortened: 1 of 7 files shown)


lib/runners/clojure.js
```diff
@@ -1,18 +1,20 @@
-var shovel = require('../shovel');
+var shovel = require('../shovel'),
+    path = require('path');
 
 module.exports.run = function run(opts, cb) {
+    var uberJar = path.resolve(__dirname, '../../jvm-runner/target/jvm-runner-0.1.1-standalone.jar');
```
javatlacati: I think groovy started failing by this commit see #175  I don't think it's by not using carmine but perhaps it's due to it's jvm-runner version [jvm-runner-0.1.2-standalone.jar](https://github.com/Codewars/codewars-runner-cli/blob/master/lib/runners/groovy.js#L5) what do you think mr @xcthulhu ? :disappointed:

javatlacati: experimenting this trough #194 :)

PR #131, opened 2015-07-17
For JavaScript: add Test.assertDeepEquals() and Test.assertNotDeepEquals()

kensnyder:
Since lodash is loaded on the server we can use the well-tested
function _.isEqual() to test deep equality. I'm not sure how to
implement in other languages since the function is hard to get exactly
right.


jhoffner:
Manually merged

PR #136, opened 2016-05-05
Updated Clojure

nrakochy:
Updated Clojure to version 1.8 and bumped jvm-runner project version from 0.1.1 to 0.1.2

### DIFF
--- a/jvm-runner/project.clj
+++ b/jvm-runner/project.clj
@@ -1,9 +1,9 @@
-(defproject jvm-runner "0.1.1"
+(defproject jvm-runner "0.1.2"
   :description "JVM Runner for codewars"
   :url "http://www.codewars.com/"
   :javac-target "1.8"
   :dependencies
-  [[org.clojure/clojure "1.6.0"]
+  [[org.clojure/clojure "1.8.0"]
    [cheshire "5.3.1"] ; JSON
    [junit/junit "4.11"] ; JUnit
    [org.codehaus.groovy/groovy-all "2.3.6"] ; Groovy
--- a/lib/runners/clojure.js
+++ b/lib/runners/clojure.js
@@ -2,7 +2,7 @@ var shovel = require('../shovel'),
     path = require('path');
 
 module.exports.run = function run(opts, cb) {
-    var uberJar = path.resolve(__dirname, '../../jvm-runner/target/jvm-runner-0.1.1-standalone.jar');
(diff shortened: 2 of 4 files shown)


jhoffner:
Awesome thanks! 

Any idea if 1.8 may cause backwards compatibility issues? We are soon rolling out the ability to target specific language versions so if there is a chance the upgrade could break older kata we may want to support both version.

nrakochy:
The [Clojure rationale](http://clojure.org/about/rationale) says that as Lisp it is not constrained by backwards compatibility. 

This PR goes from 1.6 to 1.8, and it looks like [1.7](https://github.com/clojure/clojure/blob/master/changes.md#seqs-on-java-iterators-that-return-the-same-mutating-object) has some incompatibility issues with Clojure sequences and mutating Java iterators. 

Not sure on the impliciations for backwards compatibility.

jhoffner:
OK good to know. I know nothing about Clojure. Its interesting that you didn't need to update the actual docker image to install Clojure. That means that we can target 1.6 and 1.8 within the same docker image. 

What would you recommend the best way to support multiple versions of the jvm-runner project, so that we can target both versions. Can we somehow support multiple `project.clj` files?

nrakochy:
The way to manage multiple versions can be handled through the project.clj file or with multiple profile files. You can see that [here](https://github.com/technomancy/leiningen/blob/2.3.4/doc/PROFILES.md#merging). This would probably mean a bit more scripting for firing up the separate uberjar processes as well.

I was actually hoping that I could learn Clojure via kata, so I do not know best practices for versioning and would need to defer to someone else.

By the way, Codewars is great. It has definitely been helpful. Thanks for your work.

jhoffner:
> This would probably mean a bit more scripting for firing up the separate uberjar processes as well
Good point. I'll update to 1.8 and verify each kata to make sure backwards compatibility issues aren't a problem before going down that path. 

> Codewars is great
Thanks. I'm glad you are getting benefit from it.

egri-nagy:
Just wondering what happened after the merge. codewars.com still has only Clojure 1.6.

I really miss transducers, for example, they often lead to elegant solutions.

But surely, I agree, codewars is great! :)

kazk:
@egri-nagy I just tried `(println *clojure-version*)` in Kumite and apparently we have 1.8.0 :/
image

@jhoffner `#versions` dropdown needs update.
image

PR #147, opened 2016-05-19
Avoid DD + empty string

javatlacati:
Preventing failure.getMessage to show empty message.
Code runs faster if you define a variable value only once ( final reserved word constraints compiler  to define only once in case of future modifications).
I haven't runned so please review on any branch first :$

### DIFF
--- a/jvm-runner/src/java/codewars/java/CwRunListener.java
+++ b/jvm-runner/src/java/codewars/java/CwRunListener.java
@@ -7,21 +7,18 @@ import org.junit.runner.Description;
 public class CwRunListener extends RunListener
 {
     private boolean failed;
-    public void testFailure(Failure failure)
+    public void testFailure(final Failure failure)
     {
         failed = true;
-        String msg = failure.getMessage();
-        if (msg == null) {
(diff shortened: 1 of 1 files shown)


jvm-runner/src/java/codewars/java/CwRunListener.java
```diff
@@ -7,34 +7,27 @@
...
     }
-    private static String formatMessage(String s)
+    private static String formatMessage(final String s)
     {
-        if(s == null){
-            s = "";
-        }
-
-        return s.replaceAll("\n", "<:LF:>");
+        return s == null?"":s.replaceAll("\n", "<:LF:>");
```
OverZealous: This seems like an unnecessary change.  The same result is rendered, it's just harder to see the logic now.

OverZealous: I don't know that `final` makes that much of a difference, but if it does, 2 returns would be preferable IMO.

javatlacati: But it's rendered faster :smile:

PR #148, opened 2016-05-19
Add CW-specific PHP Testing Framework

DonaldKellett:
Hello @jhoffner ,

I have just added a CW-specific PHP testing framework which outputs the custom format as required in most cases.  I have also added some basic examples in php.yml (I created a new file) and I believe I have made the PHP runner run my code examples as well as what was already there.  Please provide feedback if the testing framework still does not meet the requirements and please merge the pull request if the testing framework qualifies.

Thank you very much. :smile:

Cheers,
DonaldKellett (donaldsebleung)


frameworks/php/cw-2.php
```diff
@@ -0,0 +1,167 @@
...
+    }
+    protected function display($value) {
+      if ($value === true) return "true";
+      if ($value === false) return "false";
+      if ($value === NULL) return "NULL";
+      if (is_string($value)) return '"' . htmlspecialchars($value) . '"';
+      if (!is_array($value)) return $value;
+      if (count($value) === 0) return "array()";
+      $result = "array(";
+      $result .= "<div style='margin-left:20px'>";
```
jhoffner: Looks like we are still outputting HTML in some cases?

jhoffner:
Looks good @DonaldKellett. Did you run the test cases though? It looks like the actual integration component is missing (`/lib/runners/php.js`).

frameworks/php/cw-2.php
DonaldKellett: That is true but only when necessary.  As you will have already seen by reading my code, for key operations such as "describe" and "it" blocks, along with passes, fails and errors, only the Codewars output format is outputted.  The only cases where raw HTML is displayed is when my testing framework tries to display actual values as they look in PHP code (e.g. PHP arrays).  As stated in your `README.md` the main purpose of using the Codewars output format is as below:

>The format choosen was originally done so that at any point in time the program could exit while still having readable data.

For the current testing framework I deliberately chose to output HTML *only at places where the code is guaranteed not to unexpectedly exit* so I do not think it is possible that the program exits while outputting HTML leaving unclosed HTML tags.

Your documentation also further states that:

>With the current format there is nothing stopping you from outputing HTML, JS canvas code, etc in order to create a rich and even interactive test result output.

Hopefully my improved version of the PHP testing framework does not output too much HTML :smile:

Cheers,
donaldsebleung

DonaldKellett:
Dear @jhoffner ,

May you please elaborate on the statement you just made?

>Looks good @DonaldKellett. Did you run the test cases though? It looks like the actual integration component is missing (`/lib/runners/php.js`).

As far as I know, I added the following line of code to `test/runners/php_spec.js`:

```js
runner.assertCodeExamples('php');
```

At the time of writing, I have little to no clue what it actually does.  I'm assuming that this line of code tries to run the code examples in `examples/php.yml`?  If not, may you please include a generic version of what additional lines of code is expected in `test/runners/php_spec.js` or `lib/runners/php.js` in order to test my PHP testing framework effectively (e.g. do I have to `require('/lib/runners/php.js')` in `test/runners/php_spec.js`)?

Thank you very much for your time and help :smile:

Cheers,
donaldsebleung

frameworks/php/cw-2.php
jhoffner: True, the documentation does say that HTML is meant to be outputted, but the point there was that it could be outputted by the actual test cases deliberately, not by the test framework itself. We can leave this in there for now and once I am able to see what your output looks like maybe I'll see why its needed. I'm assuming you have some sort of fancy special output. Though to be clear, no other test framework/language on CW uses HTML in its test output (unless coded by kata author themselves).

jhoffner:
Sure. So if you go to https://github.com/Codewars/codewars-runner-cli/blob/master/lib/runners/php.js#L28 you will see the php runner. This is the code that would read in the `testFramwork` option and decide how to execute the code. Since you have not added anything here, the runner doesn't know about your framework - so at this point its basically just a file sitting in the repo that doesn't do anything. You would have to update `php.js` to integrate your code into either a newly outputted file or a concatenated string, and then run the PHP executable. You will see a line in there that is `run({name: 'php', 'args': ['-r', code]});`. Thats essentially doing this within the shell `php -r "your code"`. 

```
runner.assertCodeExamples('php');
```

You are correct, this will run your examples. Though like I said, without integrating your code into `php.js` they will all fail since  your framework isn't being utilized yet. 

Thanks for all of your effort. Its appreciated. If you don't have time to setup Docker and finish the integration I'll try to get to it tomorrow. I'm not a php guy but I think enough is done that I can figure it out.

DonaldKellett:
Thank you very much for the info and reply @jhoffner but I am still not very sure what exactly I have to add or modify in the `code` array in `php.js` to make it all work.  For example, to make `php.js` recognize my PHP testing framework, do I add a string to the `code` array containing the relative URL/path to my PHP testing framework, or do I add the PHP code to the `code` array as a string (or is it neither which is probably the case, I guess)?

I will try my best to integrate my PHP testing framework to the code runner today but in case I do not make the deadline, may you please help me with it (as I've just installed Docker like yesterday and I'm still not 100% sure how to use it)?

Thank you very much :smile:

Cheers,
donaldsebleung

jhoffner:
@DonaldKellett I'm still hoping to get some time to look into assisting you on this but I'm a bit backed up today with some other CW related changes. The best place I think for you to start is to forget about PHP for a second and reverse engineer how the other languages work. Maybe check out the `javascript.js` runner. The Ruby & Python runners would also be a good example use case, since PHP is also an interpreted language and uses a custom CW based framework, all of these languages have that in common.

javatlacati:
it passes checks now! :smile:

DonaldKellett:
Hi @jhoffner may you please help me with the PHP integration if possible? :smile:  It's been over a week already.

I apologise for not being able to study the code in the runners adequately to commit the correct changes required to properly integrate my PHP testing framework into the Codewars Runner because (1) I've been preparing for my exams lately so I don't have a lot of time on my hands (I'm just a Year 12 student studying my A-Levels who happens to know *a bit* of computer programming), (2) the code in the runners seem quite complicated to me and the runners in each language differ from each other quite a lot (for example, I do not recall an array call `code` passed into the runner for Ruby and Python) and (3) I'm just a novice developer after all (despite my "3kyu" in my Codewars Profile - I gradually earned that by completing *hundreds* of **8kyu** and **7kyu** Kata).

So, any help would be much appreciated.

Cheers,
donaldsebleung

javatlacati:
is `htmlspecialchars` better than a `xmp` tag?

jhoffner:
I'm merging now, I will look into further completing the functionality as soon as I can get to it.

DonaldKellett:
Thanks a lot @jhoffner :smile:

>is `htmlspecialchars` better than a `xmp` tag?

@javatlacati I'm not sure what you mean, may you please elaborate on your statement?

javatlacati:
<xmp><pre>code</pre></xmp>

outputs

    <pre>code</pre>

PR #154, opened 2016-05-21
Feature/dartlang

lteacher:
Hello,

This pull request includes basic [dart](https://www.dartlang.org/) language support. 

- Code running and setups are supported. 
- Test cases run via [dart test package](https://github.com/dart-lang/test). Formatting is hopefully correct?

Thanks

### DIFF
--- a/lib/runners/dart.js
+++ b/lib/runners/dart.js
@@ -53,7 +53,7 @@ var _createSolution = function(opts) {
   // that are not provided in setup
   let code = `
     ${opts.setup ? opts.setup : ''}
-    var _zimportblockstatement1;
+
     ${opts.code}
   `;
 
@@ -68,11 +68,12 @@ var _createSpec = function(opts) {
     import 'package:matcher/matcher.dart';
     ${opts.setup ? opts.setup : ''}
 
+    ${opts.code}
+
     main() {
       ${opts.fixture}
     }
 
-    ${opts.code}
   `;
 
   return _writeCode(code);
--- a/lib/runners/dart.js
+++ b/lib/runners/dart.js
@@ -49,11 +49,11 @@ module.exports.run = function run(opts, cb) {
 
 // Create the solution file
 var _createSolution = function(opts) {
-  // Set the code, _block() will stop user adding imports
+  // Set the code, the random var will stop user adding imports
   // that are not provided in setup
   let code = `
     ${opts.setup ? opts.setup : ''}
-    _block() {}
+    var _zimportblockstatement1;
     ${opts.code}
   `;
 
@@ -68,13 +68,11 @@ var _createSpec = function(opts) {
     import 'package:matcher/matcher.dart';
     ${opts.setup ? opts.setup : ''}
 
-    _block() {}
-
-    ${opts.code}
-
     main() {
       ${opts.fixture}
     }
+
+    ${opts.code}
   `;
 
   return _writeCode(code);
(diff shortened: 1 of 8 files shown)


lteacher:
Hmm I preemptively merged changes from master which is failing.. that might have been a mistake

lteacher:
Sorry but somewhere it looks like the build is failing from some other changes and is failing on master. I thought it was due to a merge that I did from master so I fixed all that stuff up but I guess the error was there from before then and I didn't notice it.

lteacher:
I was trolling around and saw something that reminded me that I forgot to add example snippets! doh...  

Cant add them right now but will do in a number of hrs.

lib/runners/dart.js
```diff
@@ -0,0 +1,171 @@
...
+        test.runTime = test.end - test.start;
+      }
+    }
+  });
+
+  return tests;
+}
+
+// Filter
+var _isUsefulLine = function(line) {
```
lteacher: Not used, remove.

lib/runners/dart.js
```diff
@@ -0,0 +1,171 @@
...
+  });
+};
+
+// Create the solution file
+var _createSolution = function(opts) {
+  // Set the code, _block() will stop user adding imports
+  // that are not provided in setup
+  let code = `
+    ${opts.setup ? opts.setup : ''}
+    _block() {}
```
lteacher: This does actually stop users creating a function called `_block`... I think that the idea is required otherwise they can put any imports they want in their code if its not one with tests. Hmmmm... Maybe a more abstract name that no one would ever use?

lib/runners/dart.js
```diff
@@ -0,0 +1,171 @@
...
+
+// Create the spec
+var _createSpec = function(opts) {
+  // Includes test suites
+  let code = `
+    import 'package:test/test.dart';
+    import 'package:matcher/matcher.dart';
+    ${opts.setup ? opts.setup : ''}
+
+    _block() {}
```
lteacher: If the ${opts.code} is simply moved below the `main() {}` function, which should be fine, then this isn't even required actually so then remove it.

lib/runners/dart.js
jhoffner: So you are intentionally trying to stop users from importing their own libraries. hmmm. If we were to do that here, we should do that everywhere - which we don't. Is there a way to make an optional method that kata authors can call which will block libraries, on a per challenge basis?

lteacher: Oh hmm I must have dreamt up this requirement when I was looking at the `setup` part. In that case probably that should be changed to allow it, but its just that when something is added in setup it will block that possibility because the imports must be at the top else its an error. So if an author puts some variable or any other functions etc into the setup no user will be able to add imports. 

Probably thats not a big deal. If the convention is to not do this then I should remove this and then switch the spec one up up again to put the code section back above main.

jhoffner:
This looks great @lteacher! I'm going to merge it in and test things hopefully later on today or tomorrow.

lteacher:
Fantastic. Let me know any issues.. am away for 2 wks from friday and wont be able to make changes from then just fyi in case

jhoffner:
Hey @lteacher. The Dart integration is live within the kumite section of the site. I don't know much about Dart but I believe in my limited testing it seemed like there were a few kinks to work out (but now I forget). I wanted to let you know so that you can start to play around with it [here](https://www.codewars.com/kumite?language=dart) and get some testing done.

lteacher:
Ok good timing because I just got back so i will look at it

Issue #175, opened 2016-06-01
Add Groovy Support


javatlacati:
When I try to create a new Kumite with the language groovy the following message is shown:

> A server error has occurred

Take for example [this Kata](http://www.codewars.com/kumite/574f1b3014b9ad877b0017af?sel=574f1b3014b9ad877b0017af) that is an exact clone of [this](http://www.codewars.com/kumite/56adafd80e62cf13a900004d?sel=56adafd80e62cf13a900004d).

javatlacati:
[groovy is not implemented](https://travis-ci.org/Codewars/codewars-runner-cli/builds/136000466#L5123)

PR #176, opened 2016-06-01
Can you please try this?

javatlacati:
Can't test it since Groovy runner seems to be down :cry:

Tryed this on tutorialspoint site, sending a file with the fixture first and below the code.

### DIFF
--- a/jvm-runner/test/codewars/runners/groovy_test.clj
+++ b/jvm-runner/test/codewars/runners/groovy_test.clj
@@ -12,6 +12,17 @@
         :code "1 + 1"})
       (is (= 2 (-main))))))
 
+(deftest groovy-basic
+  (testing "-main can handle a very basic groovy code and fixture"
+    (with-in-str
+      (json/generate-string
+       {:language "groovy"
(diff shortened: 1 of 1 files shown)


jhoffner:
Oh man. That JVM runner thing scares me (I don't code in Clojure). I'll see what I can do. I'm going to be looking into more CR stuff later on today hopefully.

javatlacati:
Yes... but if you could make the JVM runner to run this clojure code you will have instantly groovy support :dancers: 

Even if thois doesn't run Groovy has it's own assert function so you just run two programs instead of one and again you can get easily the fixture capability :smiley_cat:

jhoffner:
I'm not sure what is happening here. Did you check in a test case for functionality that you would like to be there but isn't, or for functionality that you believe to already be there but wasn't able to test?

javatlacati:
Please check  issue #175. Since groovy runner seems to be down you can't test the code. Once fixed hopefully you'll be able to run the code from [This Kata](http://www.codewars.com/kumite/574f1554f4202c807000000f?sel=574f1554f4202c807000000f) if it passes then this PR should pass and you'll have a functional junit testing framework for groovy :sunglasses:

PR #190, opened 2016-06-13
Fixes testing output for failures and errors

### LINKED ISSUE #189: Dart - Errors not output in some cases [Kumite Testing]
There are some cases where the errors are not output and instead it just outputs 'undefined', such as this example which is actually missing an entire bracket from the constructor:
image

Other cases like missing semi-colons do work however depending on where they occur, there is probably an issue where the error is taken from the test output.

lteacher:
**Fixes:**
- #189 Check for `warning:` or `error:` to output the error text for tests flagged as in error
- #191 Include `isFailure === false` when determining if result is an error or test failure

**Other Changes:**
- Use sync file writing as per most of the other runners since there is an occasional issue where some files are not written fully before accessed. As other runners use sync write I implemented the same rather than callbacks or promisifying it
- Added test case for #189 fix
- Added test case for #191 fix

### DIFF
--- a/lib/runners/dart.js
+++ b/lib/runners/dart.js
@@ -92,11 +92,11 @@ var _writeCode = function(code) {
       if (err) reject(err);
       else {
         // Write code & close
-        fs.write(info.fd, code, err => reject(err));
-        fs.close(info.fd, err => reject(err));
+        fs.writeSync(info.fd, code);
+        fs.closeSync(info.fd);
 
         // Also write this test config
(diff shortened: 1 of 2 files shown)


lteacher:
@jhoffner I have been checking on and off to see if there is any diff on the site but it seems the old image is still running there? unless the changes are totally wrong but I don't think so, rather I guess that the changes arent deployed or that the old dart image is active. 

If you want to let me know when these changes are released then I can test again on the site.

Cheers

PR #193, opened 2016-06-14
PHP Support

Dan-Nolan:
- Adds both cw-2 ( thanks @DonaldKellett ! ) and phpunit frameworks. 

I am not a PHP developer by any means, learning as necessary, so I'm opening this up for feedback from PHP developers. I'd especially like to know if the PHPUnit implementation is desirable. 

Finishing up work on proper formatting and more specs.

### DIFF
--- a/frameworks/php/phpunit/CodewarsListener.php
+++ b/frameworks/php/phpunit/CodewarsListener.php
@@ -1,0 +1,54 @@
+<?php
+    use PHPUnit\Framework\TestCase;
+
+    class CodewarsListener implements PHPUnit_Framework_TestListener
+    {
+        public function addError(PHPUnit_Framework_Test $test, Exception $e, $time)
+        {
+            $message = preg_replace('/\n/g', '<:LF:>', $e->getMessage());
+            printf("<FAILED::>%s\n", $message);
+        }
+
+        public function addFailure(PHPUnit_Framework_Test $test, PHPUnit_Framework_AssertionFailedError $e, $time)
+        {
+            $message = preg_replace('/\n/g', '<:LF:>', $e->getMessage());
+            printf("<FAILED::>%s\n", $message);
+        }
+
+        public function addIncompleteTest(PHPUnit_Framework_Test $test, Exception $e, $time)
+        {
+            // printf("Test '%s' is incomplete.\n", $test->getName());
+        }
+
+        public function addRiskyTest(PHPUnit_Framework_Test $test, Exception $e, $time)
+        {
+            // printf("Test '%s' is deemed risky.\n", $test->getName());
+        }
+
+        public function addSkippedTest(PHPUnit_Framework_Test $test, Exception $e, $time)
+        {
+            // printf("Test '%s' has been skipped.\n", $test->getName());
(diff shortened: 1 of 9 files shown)


DonaldKellett:
I personally haven't used PHPUnit before so I am not sure, but I reckon PHPUnit would be a good choice for a TDD framework because by the looks of it, it isn't as easily hacked as my testing framework (which could be easily hacked by assigning `$test` with a new value) and probably would be familiar to most PHP developers.  It could also potentially save time and speed up the (PHP) integration into Codewars as a separate documentation is not required (you could just point them to the official PHPUnit documentation instead).

frameworks/php/phpunit/CodewarsListener.php
```diff
@@ -0,0 +1,54 @@
...
+        {
+            // printf("Test '%s' is deemed risky.\n", $test->getName());
+        }
+
+        public function addSkippedTest(PHPUnit_Framework_Test $test, Exception $e, $time)
+        {
+            // printf("Test '%s' has been skipped.\n", $test->getName());
+        }
+
+        public function startTest(PHPUnit_Framework_Test $test)
```
DonaldKellett: I'm not familiar with how PHPUnit works but if the methods `startTest`, `endTest`, `startTestSuite` and `endTestSuite` are not used outside the class anyway, I think you can safely declare such methods as `protected` instead of `public`.

frameworks/php/phpunit/CodewarsListener.php
```diff
@@ -0,0 +1,54 @@
...
+
+        public function startTestSuite(PHPUnit_Framework_TestSuite $suite)
+        {
+            printf("<DESCRIBE::>%s\n", $suite->getName());
+        }
+
+        public function endTestSuite(PHPUnit_Framework_TestSuite $suite)
+        {
+            printf("\n<COMPLETEDIN::>\n");
+        }
```
DonaldKellett: I'm not familiar with how PHPUnit works but if the methods `startTest`, `endTest`, `startTestSuite` and `endTestSuite` are not used outside the class anyway, I think you can safely declare such methods as `protected` instead of `public`.

Dan-Nolan: Good call, I'll update this

javatlacati:
http://askubuntu.com/questions/452595/cannot-add-ppa-behind-proxy-ubuntu-14-04 perhaps proxy issue

javatlacati:
@jhoffner can you take a look on this :smile:

jhoffner:
@javatlacati I will as soon as I can resolve some of the remaining high priority bugs caused by the CW infrastructure transition.

javatlacati:
ty :)

frameworks/php/phpunit/CodewarsListener.php
Dan-Nolan: Turns out these have to be public since that's the access level of the methods on `PHPUnit_Framework_TestListener`

Dan-Nolan:
This should be good to go now provided Travis cooperates

PR #194, opened 2016-06-14
downgrade groovy runner version

javatlacati:
hope it can make it work again :(


javatlacati:
version update was done right but on site can do no more groovy kumites. Will experiment a little and then close sorry for the inconveniences :$

PR #196, opened 2016-06-18
Escape stderr 

Dan-Nolan:
escapes stderr output so it's not interpreted as HTML

### DIFF
--- a/lib/shovel.js
+++ b/lib/shovel.js
@@ -3,6 +3,7 @@ var util = require('util'),
     config = require('./config'),
     os = require('os'),
     utf8 = require('utf8'),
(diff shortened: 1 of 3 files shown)


javatlacati:
I can't see why it failed log is too long O.o

I suggest you tu use the tag `<xmp>` to escape output, it can escape any HTML the `<pre>` tag too and is supported almost in every browser since is a very old tag.

jhoffner:
This seemed to have failed due to erlang getting updated somehow. I'm not sure how these file changes caused the base erlang image to get rebuilt. Maybe is a fluke Travis issue. I'm going to merge and fix locally.

PR #205, opened 2016-07-02
Resolves #201

### LINKED ISSUE #201: Dart - undefined returned for some syntax errors [Kumite Testing]
Still in some examples a syntax error causes undefined to be returned such as this example just missing the semicolon in the test fixture
image

More obvious errors do seem to return:
image

lteacher:
Adds more tests and changes message splitting regex to capture even more scenarios

### DIFF
--- a/lib/runners/dart.js
+++ b/lib/runners/dart.js
@@ -118,7 +118,10 @@ var _format = function(stdout) {
       stdout += `<COMPLETEDIN::>${test.runTime}\n`;
     };
   } catch (err) {
-    stdout = `<ERROR::>${err.split(/(error|warning): /)[2]}`;
+    // Split up the error string to discard details about the dart file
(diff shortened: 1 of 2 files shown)


lteacher:
Looks like the build failed building the base-runner with a weird issue, maybe a retry would work.

jhoffner:
Unfortunately Travis can't cache docker images which sucks. The base image normally wouldn't even be rebuilt in a normal deploy. 
It might be that a package randomly failed to install. I'm not at my computer right now to check. I'll retry it later when I get back. 

Get Outlook for iOS




On Sat, Jul 2, 2016 at 4:47 AM -0700, "Hugo Armstrong" <notifications@github.com> wrote:












Looks like the build failed building the base-runner with a weird issue, maybe a retry would work.

lteacher:
Yeah dunno it says failed getting buildpack-deps:trusty which does appear to be there thats why was thinking might have just been a temp issue a retry might go through, else no clues.

PR #129, opened 2015-05-16
Intial Rust support.

Esteth:
Pretty sure I've missed out some important pieces, but I'm not sure what. I can't find a "master" dockerfile anywhere.

PR #206, opened 2016-07-10
Feature/rust

lteacher:
Hello,

This pull request implements the Rust language runner including tests.

**Notes:** 
- I couldn't get the guy's repo for #129 pull req which has been open for a year so I just implemented it. 
- Uses the default rust test capabilities. Examples are still missing since apparently I suck at actually writing code in Rust right now.
- Perhaps it should be in the `systems.docker` file. I couldn't decide so just made it separate but it seems to be just the one command so no real need for a separate image?


jhoffner:
Awesome! Without examples we wont be able to make it official yet but I can merge it and add it as a kumite language.

PR #211, opened 2016-07-15
Added back in PHPUnit output for comparing strings

OverZealous:


OverZealous:
I probably need to add some test cases in for this, I just wanted to fix it while my mind was fresh.

OverZealous:
@jhoffner This should be ready to go, this (I think) finalizes the PHPUnit-based runner.

PR #214, opened 2016-07-18
Support for julia

codeneomatrix:
*  It is incorporated julia.docker (v.0.4.6)
*  It is incorporated julia_spec.js
*  It is incorporated julia.yml
*  be updated Test.jl to version 0.4.6 
* issue

the functions:
```
facts(()->(@fact a => 1))
facts(()->(@fact  a => 1) ,"test")
```
generates the following error when run the file julia_spec.js:
```
<DESCRIBE::>test<:LF:>\n("<ERROR::>",":a - expected: 1 actual: UndefVarError(:a)<:LF:>\\n"," [inlined code] from /runner/frameworks/julia/Test.jl:194<:LF:>\\n in anonymous at /runner/frameworks/julia/Test.jl:206<:LF:>\\n in do_fact at /runner/frameworks/julia/Test.jl:158<:LF:>\\n in anonymous at none:3<:LF:>\\n in facts at /runner/frameworks/julia/Test.jl:262<:LF:>\\n in process_options at ./client.jl:257<:LF:>\\n in _start at ./client.jl:378","<:LF:>\\n")' 
to include '<DESCRIBE::>test<:LF:>\n("<PASSED::>Test Passed","<:LF:>\\n")'
  AssertionError: expected '<DESCRIBE::>test<:LF:>\n("<ERROR::>",":a - expected: 1 actual: UndefVarError(:a)<:LF:>\\n"," [inlined code] from frameworks/julia/Test.jl:194<:LF:>\\n in anonymous at /runner/frameworks/julia/Test.jl:206<:LF:>\\n in do_fact at /runner/frameworks/julia/Test.jl:158<:LF:>\\n in anonymous at none:3<:LF:>\\n in facts at /runner/frameworks/julia/Test.jl:262<:LF:>\\n in process_options at ./client.jl:257<:LF:>\\n in _start at ./client.jl:378","<:LF:>\\n")' 
  to include '<DESCRIBE::>test<:LF:>\n("<PASSED::>Test Passed","<:LF:>\\n")'
```
on lines:
```
...
 runner.run({
                        language: 'julia', 
                        code: 'a = 1', 
                        fixture:'include("frameworks/julia/Test.jl")\nusing Test\nfacts(()->(@fact a => 1))',
                        testFramework: 'Test'
                    }, function (buffer) {
                        expect(buffer.stdout).to.equal('("<PASSED::>Test Passed","<:LF:>\\n")');
                        done();
                    });
.....
 runner.run({
                        language: 'julia', 
                        code: 'global a = 1', 
                        fixture: 'include("frameworks/julia/Test.jl")\nusing Test\nfacts(()->(@fact  a => 1) ,"test")', 
                        testFramework: 'Test'
                    }, function (buffer) {
                        expect(buffer.stdout).to.contain('<DESCRIBE::>test<:LF:>\n("<PASSED::>Test Passed","<:LF:>\\n")');
                        done();
                    });
```
the code was evaluated in julia 0.4.6:
```
julia> versioninfo()
Julia Version 0.4.6
Commit 2e358ce (2016-06-19 17:16 UTC)
Platform Info:
  System: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i3-2100 CPU @ 3.10GHz
  WORD_SIZE: 64
  BLAS: libopenblas (NO_LAPACKE DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: liblapack.so.3
  LIBM: libopenlibm
  LLVM: libLLVM-3.3
```


lteacher:
Actually I never figured out why the test frameworks are required, perhaps historically its required, but just as a side note... I just used the standard rust test framework then manipulated the stdout and it doesn't seem to be a problem. 

Actually even @jhoffner updated the runner to be friendlier for that approach [as seen in the rust runner](https://github.com/Codewars/codewars-runner-cli/blob/master/lib/runners/rust.js#L56)

Probably they can elaborate on whether there is a specific need to have a purpose built test runner or not somewhere.

jhoffner:
Thanks for the contribution @codeneomatrix. 

A few notes. `<DESCRIBE::>test<:LF:>\n("<PASSED::>Test Passed","<:LF:>\\n")` isn't the fully correct format. It should be `<DESCRIBE::>test\\n<PASSED::>Test Passed\\n`. Notice no parenthesis. 

You were also including `<:LF:>` tokens within the error output, but not replacing the line breaks (`/n`) with them, which is what they are used for. When rendering error output, any line break that is used to separate text within the same error needs to be replaced.

jhoffner:
@lteacher As far as elaborating on specific needs for purpose built test runners, I feel like I'm missing part of the conversation here so I'm not sure what is being asked, but I'll elaborate on what I think you are talking about. 

The reason why we have custom built runners for some languages (like cw-2 for Ruby & JavaScript) is that in the very beginning we were not using Docker (it was only just coming out) and had to resort to using language specific sandboxing methods. That restricted my ability to use actual testing frameworks so I just built simple ones instead. As we switched over to Docker, we expanded support for actual testing frameworks but at that point CW-2 was the standard on Codewars and so others started adding support for it as we added new languages (Python, PHP). 

At this point though moving forward all new languages should use a more official testing framework. Ideally BDD/spec style if available, to be more inline with other languages on the site but it doesn't have to be if the language doesn't have one. 

As far as purpose built runner, meaning one that renders more adequately to the output that we are expecting, we do our best to make the output render in the most useful way. @lteacher the Rust support you added was the first time a straight forward find and replace was used, and I think it worked out well for Rust because in the end the output that is shown on CW is the same thing you would see if you were programming with Rust, only slightly more visual. However the more ideal experience to have a test fail and see the error within that failure, instead of having to search for the failure info down below within the STDERR section. 

Some test frameworks don't give you much leeway in customizing their output though so in the end we have work with what we got.

test/runners/julia_spec.js
```diff
@@ -1,48 +1,43 @@
...
+    it('should handle a basic assertion', function(done) {
+      runner.run({
+        language: 'julia',
+        code: 'a = 1',
+        fixture: `
+          facts(()->(@fact a => 1))
+        `,
+        testFramework: 'Test'
+      }, function(buffer) {
+        expect(buffer.stdout).to.equal('("<PASSED::>Test Passed","<:LF:>\\n")');
```
lteacher: Apparently the format here is not correct per the comments. Since it comes from the Julia runner maybe you need to change that and adjust the test cases?

lib/runners/julia.js
```diff
@@ -17,12 +19,21 @@ module.exports.run = function run(opts, cb) {
             });
         },
         testIntegration: function (run) {
-            codeWriteSync('julia', opts.solution, juliaCodeDir);
+            codeWriteSync('julia', opts.code, juliaCodeDir, 'solution.jl', true);
+
             if (opts.setup) codeWriteSync('julia', opts.setup, juliaCodeDir);
```
lteacher: Note that this `opts.setup` is not used now because the entire `juliaCodeDir` is not being pulled in, since it didn't seem to be working as desired. Maybe you could add it to the solution which is written or to the `runCmd` OR you could get the dir reading thing to work again. It did say the -P arg was deprecated when I was looking at this which convinced me to remove it

jhoffner:
Thanks for the contribution @codeneomatrix. I plan on merging this soon I just want to leave myself some time to test first.

PR #220, opened 2016-07-27
clojure fix (HACK)

jhoffner:
Clojure was escaping line breaks, causing passed/failed tokens to be treated as non-data. This is a hacky fix since i don't understand clojure well enough to fix it properly. Also cannot determine what changed that would have caused this.


OverZealous:
I hope this isn't affecting the Java runner as well.

PR #225, opened 2016-08-01
CW-2 Improvements

jhoffner:
- cw-2 now supports async
- added redis helper and tests
- python image now includes django and mpld3


Dan-Nolan:
Awesome!

OverZealous:
Nice set of upgrades!

Issue #223, opened 2016-07-29
PHP Coderunner sometimes doesn't handle Errors / Exceptions properly


DonaldKellett:
Sometimes, when the PHP code throws a certain type of Error or Exception, the PHP code runner (or the PHPUnit testing framework, I'm not sure which one is causing the problem) does not catch or handle the error properly which causes certain tests to be skipped and the faulty solution to be accepted even when it shouldn't be.

The screenshot below is an example of this bug (the highlighted section in the screenshot are the tests that were skipped due to the uncaught Error / Exception):

![2016-07-29 11 18 52](https://cloud.githubusercontent.com/assets/7777884/17253684/57734bea-55e4-11e6-915a-1e6f2f9b7e11.png)

jhoffner:
@Dan-Nolan any thoughts on what this might be?

jhoffner:
This also seems to be the same issue as #221

OverZealous:
I've got a fix for it coming.

OverZealous:
Closed via #226

PR #226, opened 2016-08-01
Improved PHP runner output. Fixes #221 and #223

### LINKED ISSUE #221: Php kata hack
<h1>Php kata  hack</h1>

Hi there is an hack for php available f.e you try to debug your code with
echo 'something'; print('something');
if you put these debugging statements in an if block or somewhere in the code before a code will be
run incorrectly it happens that the kata passed all tests without running correctly even without an  correct solution.

Another issue is: you can't copy paste things that make it harder to programm .

Ps: please  excuse me for my bad english i learn it yet.

OverZealous:
- Fixes bug causing PHP errors to get swallowed, also swallowing the `<FAILED::>` tag
    - Was caused by overzealous blacklist, those lines are now cleaned up without filtering them out
- Adds newlines before each test result, to make sure that echoing content without a newline doesn't swallow answers

### DIFF
--- a/frameworks/php/phpunit/phpunit.xml
+++ b/frameworks/php/phpunit/phpunit.xml
@@ -19,6 +19,8 @@
      stopOnSkipped="false"
      testSuiteLoaderClass="PHPUnit_Runner_StandardTestSuiteLoader"
 
+     beStrictAboutTestsThatDoNotTestAnything="true"
+
      strict="false"
      verbose="false"
     >
--- a/test/runners/php_spec.js
+++ b/test/runners/php_spec.js
@@ -238,6 +238,34 @@ describe( 'php runner', function(){
 	            });
 	        });
 
+	        it( 'should handle console output without linewraps', function(done){
+		        runner.run({
+	                language: 'php',
+	                code: `
+	                  function double($a) {
+	                    print("this was a triumph");
+	                    return $a * 2;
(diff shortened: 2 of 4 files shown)


test/runners/php_spec.js
```diff
@@ -238,6 +238,34 @@ describe( 'php runner', function(){
...
+	                    print("this was a triumph");
+	                    return $a * 2;
+	                  }
+	                `,
+	                fixture: `
+	                    class DoubleMethod extends TestCase
+	                    {
+	                        public function testDouble() {
+	                            print("I'm making a note here, huge success");
+	                            $this->assertEquals(double(1), 2);
```
Dan-Nolan: :cake:

frameworks/php/phpunit/phpunit.xml
```diff
@@ -19,6 +19,8 @@
      stopOnSkipped="false"
      testSuiteLoaderClass="PHPUnit_Runner_StandardTestSuiteLoader"
 
+     beStrictAboutTestsThatDoNotTestAnything="true"
```
Dan-Nolan: at least they're explicit!

test/runners/php_spec.js
OverZealous: _the 🎂 is a lie_

PR #228, opened 2016-08-08
Adding gmpy2 and PyCrypto to python.docker

saltthehash:
Codewars does not have a ton of cryptography katas (especially those which relate directly to practical, modern crypto), but I hope to help change that because practical modern crypto is an extremely important topic.

gmpy2 will help with high precision arithmetic needed when performing operations on very large numbers, such as in cryptography. However, it may also come in handy in other types of katas that involve working with extremely large numbers.

PyCrypto will also give users access to basic crypto functions/algorithms that are missing in Python, since Python's built-in crypto is limited to the hashlib and hmac libraries, which is insufficient.

### DIFF
--- a/docker/python.docker
+++ b/docker/python.docker
@@ -5,10 +5,10 @@
 FROM codewars/base-runner
 
 # Install Additional Python libraries
-RUN apt-get install -y python-pip python-numpy python-scipy python-pandas python-gmpy2
+RUN apt-get install -y python-pip python-numpy python-scipy python-pandas
 
 # Install Python 3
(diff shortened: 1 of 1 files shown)


jhoffner:
Thanks @shandelman116. Looks good. Just one thing though, can you move the newly added python-gmpy2 to the end of the docker file like you did PyCrypto. The trick with docker files is that the image cache is invalidated wherever you change the file, so if you change line 8 then everything else must be reinstalled. This makes deployments slower to have to redownload the bulk of the image.

shandelman116:
Thanks for the feedback @jhoffner, appreciate it! Made the change.

PR #229, opened 2016-08-10
Ruby Exceptions Should be Escaped

Dan-Nolan:
escapes ruby exceptions in rspec and cw-2

### DIFF
--- a/frameworks/ruby/cw-2.rb
+++ b/frameworks/ruby/cw-2.rb
@@ -4,6 +4,7 @@ require 'forwardable'
 require 'delegate'
 require 'set'
 require 'uri'
+require 'cgi'
 
 NameError
 class Test
@@ -20,6 +21,14 @@ class Test
 
   class << self
 
+    def escape_html(value)
+      CGI::escapeHTML(value)
+    end 
+
+    def format_exception(msg)
+      escape_html(msg)
+    end
+
     def expect(passed = nil, message = nil, options = {}, &block)
       success = (block_given? ? block.call() : !!passed)
 
@@ -30,7 +39,7 @@ class Test
         puts "<PASSED::>#{success_msg}"
       else
         message ||= 'Value is not what was expected'
-        puts "<FAILED::>#{message}"
+        puts "<FAILED::>#{format_exception(message)}"
         if $describing
           @@failed << Test::Error.new(message)
         else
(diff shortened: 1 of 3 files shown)


jhoffner:
This shouldn't be necessary. All errors should be escaped on the front end - the CR UI handles this correctly. 
We really need to get Q switched so that it uses the CR UI

Get Outlook for iOS




On Wed, Aug 10, 2016 at 2:55 PM -0400, "Dan Nolan" <notifications@github.com> wrote:












escapes ruby exceptions in rspec and cw-2



You can view, comment on, or merge this pull request online at:


  https://github.com/Codewars/codewars-runner-cli/pull/229

Commit Summary

  escapes ruby exceptions in rspec and cw-2


File Changes

  
    M
    frameworks/ruby/cw-2.rb
    (15)
  
  
    M
    frameworks/ruby/cwrspecformatter.rb
    (9)
  
  
    M
    test/runners/ruby_spec.js
    (45)
  


Patch Links:

  https://github.com/Codewars/codewars-runner-cli/pull/229.patch
  https://github.com/Codewars/codewars-runner-cli/pull/229.diff

Dan-Nolan:
@jhoffner It breaks in CW too, objects are not shown in the CR UI

jhoffner:
Ah ok it might only be happening with <ERROR::> and not STDERR then. If that's the case I wonder if double escaping causes them to display differently. 

Get Outlook for iOS




On Wed, Aug 10, 2016 at 8:29 PM -0400, "Dan Nolan" <notifications@github.com> wrote:












@jhoffner It breaks in CW too, objects are not shown in the CR UI

jhoffner:
Or maybe it's a regression or a different issue. I know I def had the <main> tag issue fixed at one point via the UI

Get Outlook for iOS




On Wed, Aug 10, 2016 at 8:29 PM -0400, "Dan Nolan" <notifications@github.com> wrote:












@jhoffner It breaks in CW too, objects are not shown in the CR UI

Dan-Nolan:
STDERR gets HTML escaped in the CLI now, but the <ERROR::> and <FAILURE::> handles for the individual language don't in some cases. I believe Phil said he added it in PHP as well.

jhoffner:
Yeah that's what I mean. Why start escaping errors on a per language level when it's a global thing. No language AFAIKT should have a need for outputting HTML within an error 

Get Outlook for iOS




On Wed, Aug 10, 2016 at 8:41 PM -0400, "Dan Nolan" <notifications@github.com> wrote:












STDERR gets HTML escaped in the CLI now, but the ERROR:: and FAILURE:: handles for the individual language don't in some cases. I believe Phil said he added it in PHP as well.

Dan-Nolan:
In that case it sounds like we should only escape once at the UI level for errors, failures and STDERR.

jhoffner:
Agreed

Get Outlook for iOS




On Wed, Aug 10, 2016 at 8:53 PM -0400, "Dan Nolan" <notifications@github.com> wrote:












In that case it sounds like we should only escape once at the UI level for errors, failures and STDERR.

Dan-Nolan:
OK, I'm closing this. Will put that into practice for Q for now since it's breaking some challenges.

PR #233, opened 2016-08-22
OCaml unit test support (using OUnit)

spencerwi:
Add support for OCaml testing (kata) support, using OUnit for testing.

I've added the commonly-used standard-lib extensions Core and Batteries to the OCaml image. I had to do some finagling to get the environment variables right between `RUN` commands in the dockerfile. 

I also had to have some convention for where the unit tests I'm running live, so tests live in a module called, well, `Tests`, and are declared in an array called `suite` (as in the example in the yml and the spec).


lib/runners/ocaml.js
```diff
@@ -4,10 +4,46 @@ module.exports.run = function run(opts, cb) {
...
+        },
+        testIntegration: function(run) {
+            var runFixtureUsingOUnit = [
+                'open OUnit',
+                'let cw_print_test_event = function',
+                '    | EStart (name::rest) -> print_endline ("<IT::>" ^ OUnit.string_of_node name)',
+                '    | EResult result -> ',
+                '        begin match result with ',
+                '        | RSuccess _ -> print_endline ("<PASSED::>Test passed")',
+                '        | RFailure (_, err) -> print_endline ("<FAILED::>" ^ err)',
```
jhoffner: You probably will need to wrap print_endline with a function that replaces `\n` with `<:LF:>`, since otherwise multiline text will get interpreted as separate data (first line would be considered a failure, all other lines would be treated as logs).

jhoffner:
Awesome! I've been wanting to play around with ocaml for a while now. 

A few things:
- specs are failing with:
```
cp docker/ocaml.docker ./Dockerfile
cp: cannot stat â€˜docker/ocaml.dockerâ€™: No such file or directory
make: *** [ocaml] Error 1
```
- missing `ocaml.yml` examples file

spencerwi:
Whoops, I missed some files in `git add`. That's embarrassing. I'll add those, then also do the `print_endline` wrapping

jhoffner:
Great, thanks for the quick turnaround. One last thing I noticed. You are missing `runner.assertCodeExamples('ocaml');` in the spec file. Including that will make sure you test the .yml examples.

spencerwi:
Ah, awesome! I didn't know about that! I'll add it.

test/runners/ocaml_spec.js
```diff
@@ -42,6 +42,10 @@ describe( 'ocaml runner', function(){
                 expect(buffer.stdout).to.contain("<DESCRIBE::>Person\n<IT::>.greet\n<PASSED::>");
                 done();
             })
-        })
+        });
+        it('should execute the included code examples', function(done){
+            runner.assertCodeExamples('ocaml');
```
jhoffner: No need for the wrapped it. See the other languages for examples. I'm not sure if it will work inside of an `it` block.

spencerwi:
Looks like the build is timing out on Travis because of how long it takes to build all the docker images including OCaml (from installing OCaml and its dependencies). Not sure what to do about that.

jhoffner:
Awesome work on this! I'll work on getting it integrated into CW hopefully in the new few days, a bit backed up with stability updates at the moment.

spencerwi:
Hey, thanks! I'm just glad to contribute. Now that I understand the process, I may have to take a crack at some of my other favorite languages...

PR #238, opened 2016-09-15
Added Karma/PhantomJS To JS Runner

OverZealous:
- Enables inclusion of FE frameworks via `@include-external «file-path»`
- Added the core files for Angular 1.2, 1.3, 1.4, and 1.5 (only some are added by default)
- Karma (with Mocha) BDD and TDD options


OverZealous:
Crap, I don't know how to hide the Angular files from the change view, but they are making a mess of everything. 😕

PR #248, opened 2016-10-13
C integration with the Criterion tests framework

pibi:


frameworks/c/criterion.c
```diff
@@ -0,0 +1,44 @@
...
+    if (stats->passed) {
+      puts("<PASSED::>Test Passed");
+    } else {
+      printf("<FAILED::>%s\n",stats->message);
+    }
+}
+
+// when a test crashes unexpectedly
+ReportHook(TEST_CRASH)(struct criterion_test_stats *stats) {
+  puts("<FAILED::>Test crashes");
```
jhoffner: "Test Crashed" would be better. Also is there any other additional information (such as exception messages) that could be passed along?

frameworks/c/criterion.c
```diff
@@ -0,0 +1,44 @@
...
+ReportHook(PRE_TEST)(struct criterion_test *test) {
+    printf("<IT::>%s\n",test->name);
+}
+
+// when an assertion is hit
+ReportHook(ASSERT)(struct criterion_assert_stats *stats) {
+    if (stats->passed) {
+      puts("<PASSED::>Test Passed");
+    } else {
+      printf("<FAILED::>%s\n",stats->message);
```
jhoffner: We should be making sure that all messages have their line breaks `/n` replaced with `<:LF:>`.

test/runners/c_spec.js
```diff
@@ -1,77 +1,169 @@
...
-describe( 'c runner', function(){
-    describe( '.run', function(){
-        it( 'should handle basic code evaluation', function(done){
-            var solution = ['#include <stdio.h>',
-                'int main() {',
-                '    printf("I think the next language codewars should support is PL/I.  Wikipedia says it\'s still used in production to this day; presumably in nursing homes?");',
-                '}',
-                ''].join('\n');
+describe('c runner', function() {
+describe('.run', function() {
```
jhoffner: Also don't forget to add examples to `examples/c.yml` and add `runner.assertCodeExamples('c');` here, so that those examples get tested. 

The examples are used later within all of our products as example code for demonstrating how to use the language.

frameworks/c/criterion.c
pibi: Unfortunately there's no other additional information. From what I understood, Criterion "catches" runtime errors (e.g. dereferencing a null pointer) and C doesn't support exception.

jhoffner:
I'm going to go ahead and merge, so that I can build locally and test. Makes sense about the C error information. The other tasks you can submit in a separate PR. 

- [ ] Replace `\n` with `<:LF:>`
- [ ] Add `c.yml` examples

PR #263, opened 2016-11-04
Added the -lm flag to link the math library.

CptDemocracy:
http://stackoverflow.com/questions/16370356/how-to-use-clang-compile-c-file-with-math-h

This should potentially resolve this issue:
https://github.com/Codewars/codewars.com/issues/658


CptDemocracy:
Just a heads-up, I'm testing it right now and it looks like my "patch" didn't work. I'm going to investigate and see how I can resolve the issue. Thank you for your time and patience!

PR #264, opened 2016-11-04
Added the -lm flag to the args array in the solutionOnly function

CptDemocracy:
Added the -lm flag to the "args" array in the solutionOnly function to provide math library linkage.

Earlier, I overlooked the "args" array in the solutionOnly function and only added the "-lm" flag to the "args" array in the testIntegration function. Hope this finally resolves
https://github.com/Codewars/codewars.com/issues/658

If not, please let me know. Thank you.

### DIFF
--- a/lib/runners/c.js
+++ b/lib/runners/c.js
@@ -9,7 +9,7 @@ module.exports.run = function run(opts, cb) {
         solutionOnly: function (run) {
             var executable = path.join(dir, 'solution'),
                 solutionFile = util.codeWriteSync('c', opts.solution, dir, 'solution.c'),
-                args = ['clang-3.6', '-std=c11', solutionFile, '-o', executable];
+                args = ['clang-3.6', '-std=c11', solutionFile, '-o', executable, '-lm'];
             if (opts.setup) {
                 var setupFile = util.codeWriteSync('c', opts.setup, dir, 'setup.c');
                 args.push(setupFile);


jhoffner:
solutionOnly won't do the trick, thats only used in Kumite when TDD is turned off. I would suggest adding a test case that specifically tries to import the Math header and see if it works, and go from there.

CptDemocracy:
Understood, thank you! I have conducted a small investigation and made a few screenshots that, I hope, will help invite new ideas and recommendations on how to resolve the issue. These screenshots were taken a few minutes ago prior to the writing of this post.

The `sqrt` function of the math library is used in the examples. The calls to it presented in the examples are legal:
http://www.cplusplus.com/reference/cmath/sqrt/?kw=sqrt

1. Calling a function from the math library in the solution window. Results in a linker error: sqrtcallincompletesolutionwindow

2. Calling a function from the math library in the test cases window. Results in a linker error:
sqrtcallintestcaseswindow

3. Why I believe the `-lm` flag could resolve the issue. Let's assume a valid code. Please, notice, that we have `math.h` included:
terminalcodeexample

4. Compilation results in a linker error without the `-lm` flag:
terminalclangcompilationwithoutlmflag

5. Compilation successful with the introduction of the `-lm` flag:
terminalclangcompilationwithlmflag

Additional resources I used for guidance:
- [Clang - How to use Clang to compile a C program that uses the math.h library](http://stackoverflow.com/questions/16370356/how-to-use-clang-compile-c-file-with-math-h)
- [Why do you have to link the math library in C](http://stackoverflow.com/questions/1033898/why-do-you-have-to-link-the-math-library-in-c)
- [GCC - Why the -lm flag is needed to link the math library?](http://stackoverflow.com/questions/4606301/gcc-why-the-lm-flag-is-needed-to-link-the-math-library)

CptDemocracy:
It looks like the issue's been resolved! Thanks Jake!

PR #271, opened 2016-11-09
feature/haskell 7.10

wizzup:
Update haskell system to newer [lts-6.24](https://www.stackage.org/lts-6.24) version (compared to #270)

Please murge this to a new `feature/haskell-7.10` branch while testing #270 (2ede419) on master.

This commit contain non-trival change to codewars's haskell framework. 

I have no idea how to fully test it beside `test/runners/haskell_spec.js` on `docker build`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,13 @@
 sudo: required
 
-language: node_js
-node_js:
-  - "5.8"
+language: javascript
 
 services:
   - docker
+
+after_success:
+  - docker ps -a
+  - docker images -a
 
 script:
   - make recent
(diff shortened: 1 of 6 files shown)


jhoffner:
Awesome Wizzup thanks! Is the non-trivial change to the test framework backwards compatible? It looks like it might be to me but I don't know Haskell.

wizzup:
> Is the non-trivial change to the test framework backwards compatible?

No, if you use the old code from old docker, the framework would not compiled due to difference in function types signature among different hspec version.

But if you mean testing user-submitted code via codewars hspec wrapper in this docker, it should be ok since it pass all test in ` codewars-runner-cli/test/runners/haskell_spec.js`  with only single modification on the last test.

I would recommend setting thing like a beta `integration test` on codewars site or find the way to pull submitted codes to re-run on this docker and see if any of it failed.

I am not a Haskell expert. I might miss some thing.

PR #287, opened 2016-12-17
Added Objective-C Linux support

idofilus:
Added Objective C support
Tests will be added soon.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ target/*
 .idea/
 .lein-repl-history
 example.rb
+Dockerfile
 *.o
 *.hi
 \#*
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,6 @@ target/*
 .idea/
 .lein-repl-history
 example.rb
-Dockerfile
 *.o
 *.hi
 \#*
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,39 +1,64 @@
-# EXAMPLE USAGE:  (FIRST BUILT): docker build -t objc .
-# EXAMPLE USAGE:  docker run objc run -l objc -c "NSLog(@\"Codewars Objective C\");"
+# EXAMPLE USAGE:  docker run --rm codewars/python-runner run -l python -c "print 1+1 "
+#                 docker run --rm codewars/python-runner run -l python3 -c "print(1+1)"
 
 # Pull base image.
 FROM codewars/base-runner
 
-# Optionally mount '/objc-vol' on ephemeral storage vol for faster
-# installing of packages, compiling, etc.
-env OBJC_VOL /objc-vol
-env OBJC_GEN_DIR $OBJC_VOL/objc-gen
-env OBJC_SOURCE_DIR $OBJC_VOL/objc-source
+# Install Additional Python libraries
+RUN apt-get install -y python-pip python-numpy python-scipy python-pandas
 
-# NEW
-RUN rm /etc/apt/sources.list
-RUN echo "deb http://archive.ubuntu.com/ubuntu/ xenial main restricted" >> /etc/apt/sources.list
-RUN echo "deb-src http://archive.ubuntu.com/ubuntu/ xenial main restricted" >> /etc/apt/sources.list
-RUN echo "deb http://archive.ubuntu.com/ubuntu/ xenial-updates main restricted" >> /etc/apt/sources.list
-RUN echo "deb-src http://archive.ubuntu.com/ubuntu/ xenial-updates main restricted" >> /etc/apt/sources.list
-RUN echo "deb http://archive.ubuntu.com/ubuntu/ xenial universe" >> /etc/apt/sources.list
-RUN echo "deb-src http://archive.ubuntu.com/ubuntu/ xenial universe" >> /etc/apt/sources.list
-RUN echo "deb http://archive.ubuntu.com/ubuntu/ xenial-updates universe" >> /etc/apt/sources.list
-RUN echo "deb-src http://archive.ubuntu.com/ubuntu/ xenial-updates universe" >> /etc/apt/sources.list
-RUN echo "deb http://archive.ubuntu.com/ubuntu/ xenial-security main restricted" >> /etc/apt/sources.list
-RUN echo "deb-src http://archive.ubuntu.com/ubuntu/ xenial-security main restricted" >> /etc/apt/sources.list
-RUN echo "deb http://archive.ubuntu.com/ubuntu/ xenial-security universe" >> /etc/apt/sources.list
-RUN echo "deb-src http://archive.ubuntu.com/ubuntu/ xenial-security universe" >> /etc/apt/sources.list
+# Install Python 3
+RUN apt-get install -y python3-pip python3-numpy python3-scipy python3-pandas
 
-RUN apt-get update
-RUN apt-get upgrade -y
+# Install Packages
+RUN pip install pymongo redis
+RUN pip3 install pymongo redis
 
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -17,7 +17,6 @@ target/*
 .idea/
 .lein-repl-history
 example.rb
-Dockerfile
```
jhoffner: This shouldn't be added. You will add a docker file to the `docker` folder and add it to the `makefile`. Then you can run `make objc` to build the file

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+
+# Set environment variables
+ENV USER codewarrior
+ENV HOME /home/codewarrior
+RUN ls -la
+RUN mocha -t 5000 test/runners/objc_spec.js
+
+# timeout is a fallback in case an error with node
+# prevents it from exiting properly
+ENTRYPOINT ["timeout", "15", "node"]
```
jhoffner: See notes above, this file shouldnt be added directly but instead add an obj.docker file to the docker folder.

lib/runners/objc.js
```diff
@@ -1,37 +1,40 @@
...
-	      objcSandbox = path.resolve('frameworks', 'osx', 'objc.sb');
+    var dir = temp.mkdirSync('objc');
+    var args = [];
 
     shovel.start(opts, cb, {
         solutionOnly: function (runCode, fail) {
-            var solutionFile = util.codeWriteSync('objc', opts.code, dir, 'main.m'),
-                command = 'clang -lobjc -framework Foundation -o main ' + solutionFile;
-
-            if ("setup" in opts) {
```
jhoffner: "setup" is still needed. This value represents code that is added along with the solution. Its often just prepended to the solution code for interpreted languages, but for compiled languages it is often compiled as its own file. (See `csharp.js` for an example)

lib/runners/objc.js
```diff
@@ -1,37 +1,40 @@
...
+            args = [solutionFile, '-lgnustep-base', '-lobjc', '-o', executable];
+
+            compile(args, function (error, stdout, stderr) {
+                if (error) return fail(error, stdout, stderr);
+                opts.publish('stdout', stdout);
+                runCode({'name': executable, 'args': []});
             });
-        },
-        testIntegration: function (runCode, fail) {
-            require('./xcode').testIntegration(opts, runCode);
```
jhoffner: Just to be clear, this will need to be added back for milestone 2

jhoffner: A set of changes that are missing is updates to `obj_spec.js`. Most of that work can probably just be re-used from the OSX version, the file just needs to be cleaned up and you can probably comment out the tests that are for `testIntegration` mode.

idofilus: Yeah, I have more code which are part of the testing .. Just wanted to upload a clean version

idofilus: Sure, I have more code .. uploaded just to make it clean for the first milestone.

lib/runners/objc.js
idofilus: Yeah, I saw something like that .. Currently now Objective C don't really need additional file since it just a core header to import and the main function ..

Dockerfile
idofilus: Yeah, I also saw it. I just uploaded it that way because this is how it worked for me .. I'm on Windows .. I don't have make .. But I'll change it ..

.gitignore
idofilus: OK, I done it like this because I'm Windows .. of course this is temporary .. just wanted to send the version that I know for sure working .. But I'll make that `make` work in my OS ..

idofilus:
Not sure why, first times I built the dockers it was OK.
But later on I had so many issues with the `apt-get update` which really that what I was spending the time on. Anyways, if you having any issues with the `apt-get update` while you install the dockers let me know .. I added the fixes on the first commit. But it may be just fine with the normal way ..

jhoffner:
Awesome, changes look good. The final thing is that I just need the objc_spec file updated so that I can verify that the build works in CI. The CI environment sometimes times out if too many images are being built, so what you also have to do in order for objc to be automatically tested is to update the makefile `RECENT_CONTAINERS=` variable to be `objc` instead of what it is now (`haskell`). That will cause your image to be built and tested. 

Once that is done I'll activate the next milestone. Great work so far!

idofilus:
I tried to run it, few things tho.
First, there are some test fails in the other languages .. (aren't they completed ?)
and second, I'm having the save issues with the `apt-get update` as I mention before, this isn't for the objc just yet. I tried to run the `make`. Because of the errors I also tried to run it on Ubuntu Desktop. Same thing. I uploaded the files this way because I wanted to know if this is also happens to you, if so. I'll upload a fix, but since the other docker seems to be like it, I guess it for some reason happens only to me ? When was the last time you tried to build it from a new clnoed project ?

jhoffner:
Travis CI (which is ran each time you push changes, see [here](https://travis-ci.org/Codewars/codewars-runner-cli/builds/185314148) for last run) will build a fresh image from scratch each time. It looks like everything built. However as you can see in the logs
```
...
Removing intermediate container 230bbefe3fb9
Step 17 : RUN mocha -t 5000 test/runners/objc_spec.js
 ---> Running in 4fc44f58d627
  0 passing (3ms)
```

The objc_specs dont have any tests so that still needs to be completed.

What is the `apt-get update` error that you are getting?

idofilus:
```
Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/p/python2.7/libpython2.7_2.7.6-8ubuntu0.2_amd64.deb  404  Not Found [IP: 91.189.88.152 80]

E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/p/python2.7/libpython2.7-dev_2.7.6-8ubuntu0.2_amd64.deb  404  Not Found [IP: 91.189.88.152 80]

E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/p/python2.7/python2.7-dev_2.7.6-8ubuntu0.2_amd64.deb  404  Not Found [IP: 91.189.88.152 80]

E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?

```

Just outdated debs or something like that, I put the fixes in the first commit if you ever occur them.
About the tests, let's keep them to the 2nd milestone .. ?

jhoffner:
Yeah maybe those packages were just down for a little bit. 

As for the unit tests. As a rule I never except any PR without test cases, including my own.  However you only need to add a single test case, it should only take a few minutes to setup.

jhoffner:
@idofilus Any updates? If you are out for the holidays that fine, just wanted to make sure we get this squared away so that I can activate the next milestone. For the c_spec, I think you pretty much just have to comment out all of the existing specs except for the first two. Also it looks like you need to remove the `if (process.platform == 'darwin') {` check at the top of the spec file so that it actually runs.

idofilus:
Didn't had time in the last few days, but I'm actually on it now .. I'll update the PR in the next few hours

lib/runners/objc.js
```diff
@@ -1,37 +1,82 @@
...
+function compile(args, cb) {
+    args.unshift('gcc', '`gnustep-config --objc-flags`');
+    exec(args.join(' '), cb);
+}
+
+function getCode(opts)
+{
+    const code = [];
+
+    if (opts.format === "default")
```
jhoffner: The CLI uses the `format` param to indicate if the response should be JSON, so this conflicts.

jhoffner:
Odd that the specs still don't run. I'm not sure why. I'm going to go ahead and merge so that you can move forward but we need to get those specs actually running.

PR #295, opened 2017-01-13
Added Unit testing, improved support of Objective C

idofilus:


headers/objc/CW/CWTest.h
```diff
@@ -0,0 +1,139 @@
...
+    {
+        NSLog(@"<PASSED::>Test Passed\n");
+    }
+    else
+    {
+        NSLog(@"<PASSED::>Test Failed\n");
+    }
+}
+
++ (void) equal:(id)a withB:(id)b
```
jhoffner: This is useful for basic matching but if it needs to handle things such as comparing a string to an int, deep matching, etc it won't be as useful. We ran into this issue with the cw framework for javascript. 

In the milestone I had mentioned integrating https://github.com/specta/expecta. How much effort would it take to support expecta with the custom output format? If expecta doesn't support an easy format integration point, what you might be able to do is catch an exception (if it throws one). 

It looks like you are not using XCTest so expecta may not be supported?

idofilus:
As you know, Xcode has a different way of executing the tests. it doesn't contain 'main()' function.
Those libraries designed for Mac and Xcode, while the different approach of executing tests doesn't contains the main(). supporting it in different environment is difficult.
Most of the libraries are built on top of XCTest, the rest are just the same as XCTest when it comes to integration with Xcode.
Your request was to support those libraries, I understand. But modifying those libraries to support cross-platform or investing time on creating a custom library with all the requirements takes more time than the price of the job is.

I have two options for you, the first is that I'll stop here and I'll have just part of the second payment.
second, is that you will send me the rules of updating the custom testing framework to receive the full 2nd payment.

I enjoyed working on it, personally I really like challenges .. but this is a subject that the internet doesn't have much information about. So it requires much time effort.
When you tag a price for this job, I personally tag the time. I cannot invest more time on it because I have deadlines of other big projects ..
So let me know what you want to prefer .. one of the two options, I can't rewrite or update big library because again. it will take much more time then I can put on this.

PR #297, opened 2017-01-18
Improved randomization features for JavaScript

kazk:
- Changed `Test.randomNumber`,`Test.randomize`,`Test.sample` to use lodash.
- Fixes [`Test.randomNumber()` not being uniform](https://github.com/Codewars/codewars.com/issues/781#issuecomment-273265048).
- Added basic tests for these functions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ CONTAINERS=node dotnet jvm python ruby alt rust julia systems dart crystal ocaml
 # recent containers should be updated when adding or modifying a language, so that
 # the travis build process will test it. The process cant test all languages
 # without timing out so this is required to get passed that issue.
-RECENT_CONTAINERS=objc
+RECENT_CONTAINERS=node
 
 ALL_CONTAINERS=${CONTAINERS} base
 
--- a/frameworks/javascript/cw-2.js
+++ b/frameworks/javascript/cw-2.js
@@ -2,7 +2,8 @@
 try
 {
     var util = require('util'),
-        deepEquals = require('lodash').isEqual,
+        _ = require('lodash'),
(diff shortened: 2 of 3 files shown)


kazk:
Do I have to do something to trigger `make node`? CI checked `codewars/objc-runner` :confused:

jhoffner:
Thanks for the contribution. How exactly do these improve the functionality? I ask because if there are any chance of different outputs happening, it might break any kata that use it. The reality is people can just import lodash if they really need it.

kazk:
Sorry, *improved* wasn't the right word.
The point of this PR is to *fix* `Test.randomNumber` because its output was not uniform. I originally did this by using `Math.floor` instead of `Math.round`. Then I saw entire lodash was imported already for `_.isEqual` so I decided to use `_.random(100)`.

I also changed `Test.randomize` and `Test.sample` to use lodash because I read your comment that those functions don't use them because that part of the framework was written before lodash was supported. Did I misunderstand you?

> I ask because if there are any chance of different outputs happening, it might break any kata that use it.

As an improvement, `_.sample(collection)` and `_.shuffle(collection)` also works for objects other than `Array` as well. But checks if `collection` is an Array, so this can happen:

```javascript
const objWithLength = {length: 3, 0: 'a', 1: 'b', 2: 'c'};
Test.sample(objWithLength); //=> one of 'a','b','c'
_.sample(objWithLength);    //=> one of 3,'a','b','c'
```

Do you think this will be a problem?

> The reality is people can just import lodash if they really need it.

I agree. I assumed the reason of having them under `Test` was simply for convenience.

kazk:
Closing this. See #308.

PR #298, opened 2017-01-20
Fixes #259 - cheating by reassigning 'global'

### LINKED ISSUE #259: [JavaScript] Test framework cheat
`global` variable can be reassigned inside solution and won't be restored after it has executed. Therefore it is possible to create a copy of original 'global' and 'Test' which effectively unfreezes them and allows to override author's test case.

Following code allows to submit pretty much any JavaScript Kata without actual solution (i. e. https://www.codewars.com/kata/reviews/5811e23a73292561ab0000b5/groups/581a53283280d6d662001037):

```
global = Object.assign({}, global)
global.Test = Object.assign({}, Test)

var od = global.Test.describe;
global.Test.describe = function(){
   return od("Fake test suite", function(){
     Test.it('fake test', function(){
       Test.assertEquals(1,1);
     });
  });
}
```

kazk:
Fixes #259 by *detecting* reassignment of `global`.

This does not prevent reassignment, but prevents cheating by throwing `Error` with message `"global was reassigned"`.

@jhoffner, do you think throwing error is valid approach?

/cc @Freywar

---

I tried `Object.freeze(global.global)` at first and found several accidental global variables. I also included fixes for those in this pull request. I decided not to use `Object.freeze` because it broke the test "should handle outputting objects with 0.10.33" and will invalidate many solutions. Not adopting suggested fix using `this` because it will fail in strict mode.

### DIFF
--- a/test/runners/javascript_spec.js
+++ b/test/runners/javascript_spec.js
@@ -585,6 +585,33 @@ describe( 'javascript runner', function(){
                         done();
                     });
             });
+
+            describe('Fix #259', function() {
+                it('should prevent global reassignment cheat', function(done) {
+                    runner.run({
+                        language: 'javascript',
+                        languageVersion: '6.6.0',
+                        code: `
+                        global = Object.assign({}, global);
+                        global.Test = Object.assign({}, Test);
+                        var od = global.Test.describe;
+                        global.Test.describe = function() {
+                          return od("Fake test suite", function() {
+                            Test.it('fake test', function() {
+                              Test.expect(true);
+                            });
+                          });
+                        };
+                        `,
+                        fixture: `Test.describe('Fail', function() { Test.it('should fail', function() { Test.expect(false); }); });`,
+                        testFramework: 'cw-2'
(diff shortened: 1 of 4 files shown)


test/runners/javascript_spec.js
```diff
@@ -585,6 +585,33 @@ describe( 'javascript runner', function(){
                         done();
                     });
             });
+
+            describe('Fix #259', function() {
+                it('should prevent global reassignment', function(done) {
```
kazk: Should have named the test case: "should prevent global reassignment cheat"

Before:
```
<DESCRIBE::>Fake test suite
<IT::>fake test
<PASSED::>Test Passed
<COMPLETEDIN::>
<COMPLETEDIN::>4
```

After:
```
<ERROR::>Error: global.global was reassigned<:LF:>           at Object.handleError<:LF:>        at ContextifyScript.Script.runInThisContext<:LF:>    at Object.exports.runInThisContext<:LF:>
```

PR #317, opened 2017-02-26
Modern objc

pibi:
Mondern Objc support


jhoffner:
@pibi This still also fully supports Objective-C 2 correct?

pibi:
@jhoffner Yes, it has the same support level of clang 3.9 for all the objc variants. Please note that I kept the original tests and I added an explicit modern version to test the seamless support.

PR #319, opened 2017-03-07
UnitKit

pibi:
OBJC - UnitKit integration


jhoffner:
@pibi looks really good! 

One issue though is that we still need to support the old framework, since there is some existing content that is already using it. A `testFramework` option is passed into the runner so we could just have the runner choose which one using `simple` or `unitkit` as the accepted values.

pibi:
@jhoffner Ok, will revert the change and keep the double test framework support. Do you want also some test for that?

jhoffner:
Yes please, though just 1 or 2 basic tests is fine, we are going to phase it out ASAP. Having it in here will just help us smooth the transition.

pibi:
@jhoffner done with `cw` and `unitkit` as names of the frameworks.

lib/runners/objc.js
```diff
@@ -105,18 +174,35 @@ module.exports.run = function run(opts, cb) {
...
             if (buffer.stderr.replace(/[ \n]/g, '') == '') {
                 buffer.stderr = '';
             }
+        },
+        sanitizeStdErr: function(error)
+        {
+            error = error || ''
+            return error.replace(/clang.*gnustep-config.*--base-libs.\t*/g, '')
+                        .replace(/Error: Command failed:/g, '')
+                        .replace(/\/home.*(solution\.m|solution)[:0-9]*/g, '')
```
jhoffner: Does this remove useful debugging information or was it confusing?

pibi: this removes the filenames and line number from the error messages because they are basically useless as they are referred to files that the user cannot access. 
The same logic is applied on the C runner

jhoffner: OK sounds good 👍

PR #323, opened 2017-03-10
Refactor lib/{config,copydir,util,shovel}.js

kazk:
Fix some inconsistencies, mainly whitespace. No logic changes.

- lib/config.js
  - Fix indentation

- lib/copydir.js
  - Fix inconsistent whitespace
  - Move default `filter` definition inside the constructor

- lib/util.js
  - Remove unused module
  - Fix inconsistencies
  - Simplify some expressions

- lib/shovel.js
  - Remove unused modules
  - Fix inconsistent whitespace
  - Fix inconsistent braces/semicolons

### DIFF
--- a/lib/shovel.js
+++ b/lib/shovel.js
@@ -1,13 +1,10 @@
 var util = require('./util'),
-    spawn = require('child_process').spawn,
-    spawnSync = require('child_process').spawnSync,
+    child_process = require('child_process'),
+    spawn = child_process.spawn,
+    spawnSync = child_process.spawnSync,
     config = require('./config'),
-    os = require('os'),
-    escapeHtml = require('escape-html'),
     temp = require('temp'),
     services = require('./services'),
-    Promise = require('bluebird'),
-    inspect = require('util').inspect,
     options = require('./options');
 
 // used for indicating an error which is related to user code and not an application error.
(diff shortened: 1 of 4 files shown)


lib/util.js
```diff
@@ -13,28 +12,26 @@ module.exports.exec = function (command, opts, cb) {
...
-        }
-        else {
+        } else {
             // if (stdout) console.log(stdout);
             if (stderr) {
-                if (cb.length == 1) cb(stderr)
+                if (cb.length == 1) cb(stderr);
                 else throw new Error(stderr);
+            } else if (error !== null) {
+                if (cb.length == 1) cb(error, stdout, stderr);
```
kazk:
```javascript
if (cb.length == 1) cb(error, stdout, stderr);
```

Why is this line checking `cb.length == 1` and passing `error, stdout, stderr`?

lib/util.js
```diff
@@ -13,28 +12,26 @@ module.exports.exec = function (command, opts, cb) {
 
     opts = opts || {};
 
-    function handleOutput (error, stdout, stderr) {
+    function handleOutput(error, stdout, stderr) {
         // wraps up any errors into a nice package and returns the json value.
-        if (opts.handleError && ((error && error != '') || (stderr && stderr != ''))) {
```
kazk: `error != ''` is unnecessary because `''` won't reach there.

lib/util.js
jhoffner: Good question, I have no idea. This entire file needs to be rewritten IMO. We have so many different custom versions of `exec` floating around. I still need to do a full audit.

lib/shovel.js
```diff
@@ -117,8 +106,7 @@ function setupFiles(opts, strategies) {
     if (opts.files) {
         if (strategies.files) {
             strategies.files();
-        }
-        else {
+        } else {
```
jhoffner: I'm more of a 
```
if () {
}
else {
}
```
kind of guy (easier to scan since all keywords are on the same vertical line) but we can go with this for now ;)

jhoffner: This is also probably because I'm a Ruby guy :)

jhoffner:
Looks good. Thanks!

lib/shovel.js
kazk: Sorry about that. I thought the majority of the code had `} else {` and misunderstood your comment about `{` falling to the next line.

PR #337, opened 2017-03-23
Minor style changes

javatlacati:
I always change the code I read hope you don't mind :smile:

### DIFF
--- a/frameworks/javascript/cw-2.js
+++ b/frameworks/javascript/cw-2.js
@@ -41,7 +41,9 @@ try
     process.on('uncaughtException', function (err) {
         if (async) {
             Test.handleError(err);
-            if (asyncDone) asyncDone();
+            if (asyncDone){
+              asyncDone();
+            }
         }
     });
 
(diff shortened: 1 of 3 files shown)


kazk:
For JavaScript files, I've been looking into and working on to propose `esformatter`, `ESLint` to keep the codebase consistent. Can we wait?
I personally don't see any gain in readability/maintainability by adding braces for short, single statements like that. But if we agree on this, we can configure the rules to enforce this across the codebase and not only few files in this PR.

---

I hate Yoda conditions 🙀

kazk:
This is why I prefer using tools for these things.

[frameworks/javascript/cw-2.js#L220](https://github.com/javatlacati/codewars-runner-cli/blob/9c0a3560f54aaac726d6409b4da143f2e8102e77/frameworks/javascript/cw-2.js#L220)
```js
if (timeout) clearTimeout(timeout)
```

[frameworks/javascript/cw-2.js#L231](https://github.com/javatlacati/codewars-runner-cli/blob/9c0a3560f54aaac726d6409b4da143f2e8102e77/frameworks/javascript/cw-2.js#L231)
```js
if (asyncIt) describeNext();
```

[jvm-runner/src/java/codewars/java/CwRunListener.java#L37](https://github.com/javatlacati/codewars-runner-cli/blob/9c0a3560f54aaac726d6409b4da143f2e8102e77/jvm-runner/src/java/codewars/java/CwRunListener.java#L37)
```java
if(ex == null){
```

kazk:
Also, `RECENT_CONTAINERS` in Makefile needs to be changed because Travis checked `go-runner`.

PR #343, opened 2017-03-28
Fix Inconsistent Coding Style

### LINKED ISSUE #342: ESLint 'brace-style'
I forgot to fix the indentation of comments and I started to work on this.
When I opened [frameworks/javascript/cw-2.js](https://github.com/Codewars/codewars-runner-cli/blob/321c62e176e4b0e031e355ffc973a5e23c6d6b18/frameworks/javascript/cw-2.js), I found that it looks terrible because of inconsistent braces.
This can be fixed using `eslint --fix` with [brace-style](http://eslint.org/docs/rules/brace-style) rule, but which rule should we use?

The most common style is ["1tbs"](http://eslint.org/docs/rules/brace-style#1tbs):

```js
if (foo) {
  bar();
} else {
  baz();
}
```

@jhoffner [prefers](https://github.com/Codewars/codewars-runner-cli/pull/323#discussion_r105473359) `else` on its own line and there's a rule for that called ["stroustrup"](http://eslint.org/docs/rules/brace-style#stroustrup):

```js
if (foo) {
  bar();
}
else {
  baz();
}
```

Another one is called ["allman"](http://eslint.org/docs/rules/brace-style#allman):

```js
if (foo)
{
  bar();
}
else
{
  baz();
}
```

I think this should be decided by Codewars team to avoid holy war.
@jhoffner, @OverZealous, @Dan-Nolan, @nathandoctor

---

(I should have asked this within "Start using ESLint #338" before merge :()

kazk:
Please review individual commits for the changes applied.

Closes #342.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -100,7 +100,7 @@ module.exports.writeFileSync = function(dir, fileName, content, overwrite) {
   mkdirParentSync(path.dirname(fileName));
   fs.writeFileSync(fileName, content);
   return fileName;
-}
+};
 
 // writes multiple files. Also allows you to transform the content before it gets written. Files should be an object
 // with keys mapping to file names and values mapping to file content.
@@ -115,4 +115,4 @@ module.exports.writeFilesSync = function(dir, files, overwrite, transform) {
       util.writeFileSync(dir, fileName, content, overwrite);
     });
   }
-}
+};
--- a/lib/util.js
+++ b/lib/util.js
@@ -4,7 +4,7 @@ var fs = require('fs'),
   child_process = require('child_process'),
   util = module.exports;
 
-module.exports.exec = function (command, opts, cb) {
+module.exports.exec = function(command, opts, cb) {
   if (!cb) {
     cb = opts;
     opts = {};
@@ -12,7 +12,7 @@ module.exports.exec = function (command, opts, cb) {
 
   opts = opts || {};
 
-  function handleOutput (error, stdout, stderr) {
+  function handleOutput(error, stdout, stderr) {
         // wraps up any errors into a nice package and returns the json value.
     if (opts.handleError && (error || stderr)) {
       cb({
@@ -73,7 +73,7 @@ function codeFileName(language, code, defaultFileName) {
   }
 }
 
-module.exports.codeWriteSync = function (language, code, codeDir, defaultFileName, overwrite) {
+module.exports.codeWriteSync = function(language, code, codeDir, defaultFileName, overwrite) {
   var fileName = codeFileName(language, code, defaultFileName);
   if (!(typeof fileName == 'string' || fileName instanceof String)) {
     throw new Error("Could not determine valid name from code:\n\n" + code);
@@ -106,7 +106,7 @@ module.exports.writeFileSync = function(dir, fileName, content, overwrite) {
 // with keys mapping to file names and values mapping to file content.
 module.exports.writeFilesSync = function(dir, files, overwrite, transform) {
   if (files) {
-    Object.keys(files).forEach(function (fileName) {
+    Object.keys(files).forEach(function(fileName) {
       var content = files[fileName];
       if (transform) {
         content = transform(fileName, content);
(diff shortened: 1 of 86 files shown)


kazk:
@jhoffner what do you think of this rule?

['no-lonely-if'](http://eslint.org/docs/rules/no-lonely-if)

```js
// turns this
if (foo) {
}
else {
  if (bar) {
  }
}

// into
if (foo) {
}
else if (bar) {
}
```

---

There's also a rule to always have braces, including a simple single statements (#337).

['curly'](http://eslint.org/docs/rules/curly)

```js
if (foo) foo++;
// to
if (foo) {
  foo++;
}
```

kazk:
['no-else-return'](http://eslint.org/docs/rules/no-else-return)?

```js
function foo() {
  if (x) {
    return y;
  }
  else {
    return z;
  }
}

// to
function foo() {
  if (x) {
    return y;
  }
  return z;
}
```

jhoffner:
Both of those rules look good to me.

kazk:
Do you mean 'no-lonely-if' and 'no-else-return'?

'curly' has some complicated options that I haven't been able to look into.

lib/util.js
```diff
@@ -21,20 +21,20 @@ module.exports.exec = function(command, opts, cb) {
         exitCode: error && error.code,
         exitSignal: error && error.signal
       });
+      return;
```
kazk: Added this `return` to help `eslint`. This is how it fixed without it, I think the comment is interfering from recognizing `else if`.

```js
if (opts.handleError && (error || stderr)) {
  cb({
    stdout: stdout,
    stderr: [stderr, error].filter(v => !!v).join(''),
    exitCode: error && error.code,
    exitSignal: error && error.signal
  });
} else {
  // if (stdout) console.log(stdout);
  if (stderr) {
    if (cb.length == 1) cb(stderr);
    else throw new Error(stderr);
  }
  else if (error !== null) {
    if (cb.length == 1) cb(error, stdout, stderr);
    else throw error;
  }
  else {
    cb();
  }
}
```

jhoffner:
One last possible tweak before I merge this. We may want to keep a space between parenths when declaring a function. 

```
function foo () {
}

foo();

// instead of 
function foo() {

}
```

I actually don't really care much either way, but when building out the new CodeRunner UI I went mostly with the eslint defaults that VueJS uses, and this was one of them. It would be nice to keep these two projects similar. 

For reference, here is the eslintrc file for that project, with some of my modifications

```javascript
module.exports = {
  root: true,
  parser: 'babel-eslint',
  parserOptions: {
    sourceType: 'module'
  },
  extends: 'standard',
  // required to lint *.vue files
  plugins: [
    'html'
  ],
  // add your custom rules here
  'rules': {
    // allow paren-less arrow functions
    'arrow-parens': 0,
    // allow async-await
    'generator-star-spacing': 0,
    // allow debugger during development
    'no-debugger': process.env.NODE_ENV === 'production' ? 2 : 0,
    'spaced-comment': 0,
    'no-multiple-empty-lines': 0,
    'brace-style': 0,
    'valid-typeof': 0,
    'one-var': 0,
    'indent': 0,
    'quotes': 0,
    'comma-dangle': 0,
    'no-throw-literal': 0,
    'no-multi-spaces': 0,
    'padded-blocks': 0,
    'key-spacing': 0,
  }
}
```

jhoffner:
With this said, I constantly forget to add the space and end up having to go back and fix it, so if anyone has strong opinions against having the space, I'm fine with just changing the eslint for the UI project instead.

kazk:
I think that space after the function name comes from 'standard' rule. (I thought it was controlled by ['space-before-function-paren'](http://eslint.org/docs/rules/space-before-function-paren) rule and I didn't see that in your configuration.)

To be honest, I don't like that space for function declaration. I'll most likely forget about it too. But I don't have *strong* opinion against it, I'm probably just not used to it.
I'm guessing that this rule exists to make declaration vs invocation more clear.

jhoffner:
We don't use the space within the Qualified codebase either so I'll just change it for the UI project.

PR #347, opened 2017-03-31
[RFC][WIP] Improved Python POC

kazk:
### Features

- No file concatenation
- Tracebacks with line numbers/file names
- Test integration using standard unittest
- Can be modified to have multiple test files
- Can be modified to test a package/project

---

```
Current                   Testing a package     Multiple test files
.                         .                     .
|-- setup.py              |-- kata              |-- kata
|-- solution.py           |   |-- __init__.py   |   |-- __init__.py
`-- test                  |   |-- setup.py      |   |-- setup.py
    |-- __init__.py       |   `-- solution.py   |   `-- solution.py
    |-- __main__.py       `-- test              `-- test
    `-- test_solution.py      |-- __init__.py       |-- __init__.py
                              |-- __main__.py       |-- __main__.py
                              `-- test_kata.py      |-- test_kata1.py
                                                    `-- test_kata2.py

All of these should be supported without much modification.
```

At first, I looked into running the test with `python -m unittest discover`, but I couldn't figure out a way to use `CodewarsTestResult` with this.
`test/__main__.py` uses `loader.discover('.')` to find all the tests in `test` package and runs them by calling `unittest.main()` with `testRunner=CodewarsTestRunner`.

---

### TODO

- Grouping tests
- Cleanup messages

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,12 @@
 HOSTNAME=codewars
 
 # Building erlang images have been suspended (frozen) until they are able to be repaired
-CONTAINERS=node dotnet jvm python ruby alt rust julia systems dart crystal ocaml swift haskell objc go lua
+CONTAINERS=node dotnet jvm python ruby alt rust julia systems dart crystal ocaml swift haskell objc go lua python-poc
 
 # recent containers should be updated when adding or modifying a language, so that
 # the travis build process will test it. The process cant test all languages
 # without timing out so this is required to get passed that issue.
-RECENT_CONTAINERS=node
+RECENT_CONTAINERS=python-poc
 
 ALL_CONTAINERS=${CONTAINERS} base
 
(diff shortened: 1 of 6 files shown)


jhoffner:
Thanks for this, this is definitely an improvement. 

We will have to identify all of the backwards incompatibility issues. Since CW uses cw-2 only right now, all kata should be unaffected, but we do have Qualified content created using unittest. 

It looks like the 2 biggest changes are that we need to `import unittest` and `import solution` to the top of the spec and update the references to use solution. There may be other scoping issues that make the change difficult. 

The best way to handle it may be to support the current system as a legacy mode, defined by adding the following to the setup code:

```
# concatenate the code blocks into one file
# @config useConcat true
```

kazk:
Superseded by #552

PR #360, opened 2017-04-01
Fix lib/runners/csharp.js

kazk:
- Remove unnecessary assignment and commented out code
- Remove unused module in csharp_spec.js

### DIFF
--- a/lib/runners/csharp.js
+++ b/lib/runners/csharp.js
@@ -90,7 +90,7 @@ module.exports.run = function run(opts, cb) {
       compile(args, function(error, stdout, stderr) {
         if (error) return fail(error, stdout, stderr);
 
-        const exe = path.join(opts.dir, 'runner.exe');
+        // const exe = path.join(opts.dir, 'runner.exe');
 
         // compile([
         //     'mcs',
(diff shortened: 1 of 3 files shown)


lib/runners/csharp.js
```diff
@@ -89,41 +89,6 @@ module.exports.run = function run(opts, cb) {
       // compile
       compile(args, function(error, stdout, stderr) {
         if (error) return fail(error, stdout, stderr);
-
-        const exe = path.join(opts.dir, 'runner.exe');
-
-        // compile([
```
jhoffner: This was a WIP. I had just left it in so that I would either remember to complete it or someone else would think to run with it. 

I actually gave up on it in frustration with dealing with C# assembly references but had hoped to get back to it soon. If we were to clean this up then we would probably just want to clean up the files it references as well, so maybe its better to just leave it in. I know people always say we can resurrect it from Git history but that rarely actually happens - more likely it will be forgotten.

kazk: Ok, I'll just comment out that `const exe = path.join(opts.dir, 'runner.exe');`.

kazk:
Maybe Travis didn't like my rebase. I'll try again, also resolving the merge conflict.

PR #366, opened 2017-04-04
ARC support

pibi:
This enables ARC support for objc


kazk:
`RECENT_CONTAINERS` needs to be changed to `objc`. Travis checked `codewars/node-runner`.

image

lib/runners/objc.js
```diff
@@ -4,11 +4,12 @@ var shovel = require('../shovel'),
   exec = require('child_process').exec;
 
 function compile(args, cb) {
-  args.unshift('clang', '`gnustep-config --objc-flags --objc-libs`');
+  args.unshift('clang',  '-fobjc-arc', '`gnustep-config --objc-flags --objc-libs`');
```
jhoffner: If ARC support can possibly cause headaches for non-ARC based solutions, we can support this as a language version. The API supports a `languageVersion` option. Perhaps the values could be 
- GNUStep (default)
- ARC (or maybe ModernC?)

Let me know if you have better suggestions

PR #369, opened 2017-04-05
Add assert(Not)FuzzyEquals for floating point comparison

DonaldKellett:
# Add assert(Not)FuzzyEquals for floating point comparison

This pull request aims to add two new `Test` methods, `Test.assertFuzzyEquals` and `Test.assertNotFuzzyEquals` to the current CW-2 JavaScript testing framework

## The Problem

Every once in a while, a relatively inexperienced Codewars Kata author authors a Kata which involves calculations resulting in floating point values.  Usually, such Kata authors will naturally assume that using `Test.assertEquals`/`Test.assertNotEquals` is sufficient for comparing two floats.  What they do **not** know is that due to floating point errors in JavaScript, the same calculation carried out in a different order may result in a *very slightly* different float which causes valid answers to be rejected by `Test.assertEquals`/`Test.assertNotEquals`.  In such cases, Moderators like me are required to intervene in such Kata by raising an Issue regarding this, which is commonly accompanied by a custom assertion method defined on the fly called `assertFuzzyEquals` which I tell the Kata author to copy-paste to their Preloaded section in order to solve the problem.  All is fine, except that every time I raise an Issue regarding floating point errors, **I have to manually type the entire custom assertion from memory** which is extremely annoying, time-consuming and error-prone.

## The Solution

Instead of me writing the exact same custom assertion method for the Kata author every time someone incorrectly uses `Test.assertEquals` to compare two floats, the addition of `Test.assertFuzzyEquals` and `Test.assertNotFuzzyEquals` as officially recognized CW-2 assertion methods makes the life of us Moderators much easier as we need only say something like “Please use `Test.assertFuzzyEquals` instead of `Test.assertEquals` to compare decimal numbers” next time an amateur Kata author makes this same mistake again.

## Credits

The entire logic for these two assertion methods are shamelessly stolen - *ahem*, borrowed - from the [conjured Codewars Codex](http://bkaestner.github.io/codewars-rules/#floating-point-tests-1) (aka the Codewars Bible) authored by @bkaestner


bkaestner:
Sidenote: How did you manage to get 29 commits for a single change? What happened there?

frameworks/javascript/cw-2.js
```diff
@@ -356,6 +356,12 @@ try {
         Test.expect(false, msg, options);
       }
     },
+    assertFuzzyEquals: function (actual, expected, msg = "") {
```
kazk: The test is failing because of `SyntaxError`:

```
/runner/frameworks/javascript/cw-2.js:359
   assertFuzzyEquals: function (actual, expected, msg = "") {
                                                       ^
SyntaxError: Unexpected token =
```

I didn't realize this until I saw this, but the test framework needs to be compatible with Node 0.10.x.

kazk:
image

Set `upstream` to `Codewars/codewars-runner-cli` and follow this [guide](https://help.github.com/articles/syncing-a-fork/) to avoid all these merge commits.

I'd also recommend making a branch for each feature/fix instead of working on `master`. I never work on `master` because undoing can be painful. I also make experimental branches locally which I can just delete if it doesn't work.

frameworks/javascript/cw-2.js
jhoffner: Also can we comment the method to indicate its fuzzy for float values. I could see it easily being confused with `assertSimilar`

bkaestner: Or one could change `assertSimilar` to use this kind of check if both types are `number`s and not `int32`.

kazk: We need tests in `test/runners/javascript_spec.js` having a `fixture` allowing inexact solution.

bkaestner:
From a quick review, the following parts are still needed:

- [ ] follow the style of the other assertions 
- [ ] make the assertions compatible to the nodejs version run by CodeWars
- [ ] provide test cases to check whether the assertions do their job

frameworks/javascript/cw-2.js
kazk: How about renaming them `assertAlmostEqual(s)`/`assertNotAlmostEqual(s)` (from Python's `unittest`)?

DonaldKellett:
Most recent commit should check the following checkboxes:

- follow the style of the other assertions
- make the assertions compatible to the nodejs version run by Codewars (default parameter value removed)

TODO: provide test cases to check whether the assertions do their job

kazk:
Will merging this PR inject all these merge commits into the repository history? I've never seen anything like this and I'm not sure.
If so, I found [a question](http://stackoverflow.com/questions/5968964/avoid-unwanted-merge-commits-and-other-commits-when-doing-pull-request-on-github) on stack overflow which might help.
If that's too much, simply doing
1. back up `cw-2.js`
2. reset everything
3. sync the fork
4. create the new branch and replace `cw-2.js` with backed up file
5. opening a new PR

is better since it's still a single file change.

DonaldKellett:
Just added commit with tests testing the functionality of `Test.assert(Not)ApproxEquals`

DonaldKellett:
> Will merging this PR inject all these merge commits into the repository history? I've never seen anything like this and I'm not sure.

Apologies @kazk I read your comment just now.  Would it still be possible to accept my pull request as long as all checks are successful this time?  If not, I can still close this pull request and try to start a new one that doesn't involve `master`.

kazk:
@DonaldKellett no need to apologize :) I'd wait @jhoffner because this might not be a problem for this repo.

kazk:
@DonaldKellett even if this is a problem, you don't need to make a new PR. `git pull --rebase upstream master` from [the answer](http://stackoverflow.com/a/5969100) for the stack overflow question worked for me.
 
Tested with

```
git fetch upstream pull/369/head:feature/js-assert-fuzzy-equals
git checkout feature/js-assert-fuzzy-equals
git pull --rebase upstream master
```

So in your case

```
git remote add upstream https://github.com/Codewars/codewars-runner-cli
git checkout master
git pull --rebase upstream master
git push --force origin master
```

should remove all that merge commits from this PR. Just be careful in case I misunderstood your setup.

DonaldKellett:
Thanks @kazk for the link to the StackOverflow question and your code example.  I just tried executing your code example and all went well except for the final line `git push --force origin master` where I assumed `origin === DonaldKellett` (I've never been very good at Terminal).  I then headed back to my GitHub Desktop app and realised that my 3 most recent commits were still there but not synced with the online servers so I manually pressed "Sync" and all the previous merge commits reappeared.  I'll try to close this pull request and create a new one and see if that solves the problem.  If not, I'll try to find another way.  Many thanks for your help and advice again :smile:

PR #386, opened 2017-04-22
Fix ESLint issues for node-runner

kazk:


lib/services.js
```diff
@@ -19,24 +19,21 @@ const startService = {
 
     return spawnAndWait(opts, '/usr/lib/postgresql/9.6/bin/pg_ctl', ['-D', '/pgdata/pg', '-w', 'restart']);
   },
-  mariadb: function(opts) {
+  mariadb: function(opts) { // eslint-disable-line no-unused-vars
```
jhoffner: It looks like we could disable this rule for arguments, using `args: 'none'` within the no-unused-vars config. I personally prefer to have all arguments listed out, especially in OS projects where a contributor is going to be less familiar with the code. It has the side-affect of not making it easy to determine if a value is actually being used or not, but has the benefit of acting like documentation, it's very convenient to know the options available to you so that when you need to jump into the code you have some reference from the start.

kazk: I didn't know about that option. I'll add it in `.eslintrc.yml` instead.

PR #393, opened 2017-04-22
Delete lib/runners/xcode.js

kazk:
Delete `lib/runners/xcode.js` and related files.
If this is still used, I'll fix the ESLint issue in `lib/runners/xcode.js`.


jhoffner:
Good call. Thanks!

PR #394, opened 2017-04-23
Fix "repository" in package.json

kazk:
Also added: "contributors", "bugs", "license".

Used GitHub API for "contributors" because `git shortlog -sn` returned 3 Jakes.

```
curl https://api.github.com/repos/codewars/codewars-runner-cli/contributors \
    | jq '[.[] | select(.login != "jhoffner") | .login]'
```

The list is sorted by the number of commits.


jhoffner:
Awesome! Thanks.

Issue #328, opened 2017-03-12
Refactor Dockerfiles


kazk:
Dockerfiles need to be refactored to reduce image sizes, follow best practices and ensure reproducibility.

### `codewars/swift-runner`
[![](https://images.microbadger.com/badges/image/codewars/swift-runner.svg)](https://microbadger.com/images/codewars/swift-runner "Get your own image badge on microbadger.com")

#### Problems
- Download size is `2.7 GB` 😱 
- Swift version not specified 😱 

#### Possible Solutions

- Download ([L33](https://github.com/Codewars/codewars-runner-cli/blob/f06cca39981b34f48d4678e0cd98eef8a5a876f8/docker/swift.docker#L33)), build ([L45](https://github.com/Codewars/codewars-runner-cli/blob/f06cca39981b34f48d4678e0cd98eef8a5a876f8/docker/swift.docker#L45)) and remove ([L49](https://github.com/Codewars/codewars-runner-cli/blob/f06cca39981b34f48d4678e0cd98eef8a5a876f8/docker/swift.docker#L49)) in the same layer.

  ```dockerfile
  # L33
  RUN git clone https://github.com/apple/swift.git && ./swift/utils/update-checkout --clone
  # 783.3 MB
  
  # L45
  RUN utils/build-script --preset-file=codewars-build-presets.ini --preset=codewars_linux_1404 install_destdir=$SWIFT_GEN_DIR installable_package=$SWIFT_GEN_DIR/swift.tar.gz
  # 944.1 MB
  
  # L49
  RUN rm -rf $SWIFT_SOURCE_DIR/*
  ```

- Find a better way to get the source at specific version. Cloning the entire history shouldn't be necessary.
  - `3.0-dev`([Language Support Status](https://github.com/Codewars/codewars-runner-cli#language-support-status))
  - [swift-3.0-RELEASE](https://github.com/apple/swift/releases/tag/swift-3.0-RELEASE) (`2016-09-13`)
  - [swift-3.0.1-RELEASE](https://github.com/apple/swift/releases/tag/swift-3.0.1-RELEASE) (`2016-10-28`)
  - [swift-3.0.2-RELEASE](https://github.com/apple/swift/releases/tag/swift-3.0.2-RELEASE) (`2016-12-13`)

## Resources

- [Best practices for writing Dockerfiles][BestPractices]
[...]

PR #399, opened 2017-04-25
Use Travis build matrix

kazk:
image

Skipping Swift because it timeouts. See my newest comment on #328.

Python is failing because `IPython` had a major version increment dropping support for older Python. Some other packages are failing too.


jhoffner:
Whoa! Awesome! 

I have a fix for python ready. Will push in a bit.

PR #400, opened 2017-04-26
Fixed Python, minor code cleanup

jhoffner:
Python package removed that was causing issues. Python versions need to be upgraded before IPython can be used (no one was using it anyway).


lib/options.js
```diff
@@ -4,6 +4,12 @@ module.exports.process = function(opts) {
   // set the default directory to run everything out of
   opts.dir = opts.dir || '/home/codewarrior';
 
+  // opts.code is an alias for opts.solution
+  if (opts.code) {
+    opts.solution = opts.code;
+    delete opts.code;
```
kazk: We can't have `delete opts.code;` yet because some language runners uses `opts.code`.

The build matrix caught this :)

image

lib/shovel.js
```diff
@@ -1,11 +1,11 @@
 var util = require('./util'),
-    child_process = require('child_process'),
```
kazk: > would have preferred 4 space var indents and indented switches but I can get used to this.


```yaml
rules:
  indent:
    - error
    - 2
    - VariableDeclarator:
        var: 2
        let: 2
        const: 3
      SwitchCase: 1
```

will do. I'll open a PR.

kazk:
Fixed this in #402 because I thought it'll be the easiest.

PR #402, opened 2017-04-26
Add options to ESLint 'indent' rule

kazk:
- Align declarations
- Indent `case`

Fixed the code to match this rule.

### DIFF
--- a/lib/runners/cpp.js
+++ b/lib/runners/cpp.js
@@ -1,11 +1,11 @@
 var shovel = require('../shovel'),
-  util = require('../util'),
-  path = require('path'),
-  temp = require('temp'),
-  exec = require('child_process').exec,
-  fs = require('fs'),
-  cppDir = path.resolve(__dirname, '..', '..', 'frameworks', 'cpp'),
-  main = fs.readFileSync(path.resolve(cppDir, 'main.cpp'));
+    util = require('../util'),
+    path = require('path'),
+    temp = require('temp'),
+    exec = require('child_process').exec,
+    fs = require('fs'),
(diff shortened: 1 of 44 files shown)


kazk:
I don't know why `systems-runner` is failing :/
image

kazk:
Looks like it's related to `opts.setup` for C++ :/

kazk:
This is not related ESLint fixes. I made a version only fixing #400 (kazk/codewars-runner-cli@72e4763803ce09101ea9efb42091735e2d168bf3) and still fails.

Before #400, the build matrix only failed for `python`.

kazk:
I think I figured this out. `lib/runners/cpp.js` uses `opts.solution` **before** `shovel` when `opts.setup` is present.

```js

  var args = [];
  if (opts.setup) {
    var setupFile = util.codeWriteSync('cpp', opts.setup, dir, 'setup.h');
    args.push(setupFile);
    opts.solution = '#include "setup.h"\n' + opts.solution;
  }

  shovel.start(opts, cb, {
```

lib/runners/cpp.js
```diff
@@ -20,6 +20,7 @@ module.exports.run = function run(opts, cb) {
   if (opts.setup) {
     var setupFile = util.codeWriteSync('cpp', opts.setup, dir, 'setup.h');
     args.push(setupFile);
+    if (opts.code) opts.solution = opts.code;
     opts.solution = '#include "setup.h"\n' + opts.solution;
```
kazk: Fails because `opts.solution` is overwritten by `opts.code` in `shovel`. Fixing this now.

kazk:
image

Issue #326, opened 2017-03-10
Replace jvm-runner with a non-clojure solution


jhoffner:
The current clojure based solution is next to impossible for the core team to maintain. We need to move off of this. 

The purpose for the jvm-runner is to improve JVM startup performance. Other existing solutions, such as GroovyServ, should be considered for the replacement.

kazk:
[GroovyServ](https://kobo.github.io/groovyserv/)
- [kobo/groovyserv](https://github.com/kobo/groovyserv)
- [FAQ](https://kobo.github.io/groovyserv/faq.html) has some info about [Drip](https://github.com/ninjudd/drip) and [Nailgun](http://www.martiansoftware.com/nailgun/)

> [Is GroovyServ dedicated for Groovy?](https://kobo.github.io/groovyserv/faq.html#is-groovyserv-dedicated-for-groovy)
> No. you can use GroovyServ for other JVM languages.
> For example, you can run a Clojure script like this:
> ```
> $ CLASSPATH=/tmp/clojure.jar groovyserver
> $ groovyclient -e 'import clojure.main;main.main(args)' -- -e "(println 'Clojure)"
> Clojure
> ```

---

[Scala Native](http://www.scala-native.org/) (pre-release)
- [scala-native/scala-native](https://github.com/scala-native/scala-native)
- [Docs](http://scala-native.readthedocs.io/en/latest/)

OverZealous:
GroovyServ is interesting to me because:

1. Groovy is pretty easy to understand, since it's much closer to traditional C-like languages
2. Groovy has great native Java-library support
3. You can use `@Grab()` to automatically download & include just about any Maven library (and some others), which are cached locally. Obviously that last part only helps for long-running connections, but it could open the door to easily downloading any library upon first run.
4. I believe you can pass in changes to the classpath _after_ warming up the JVM. This has some potential for opening the door to include custom JARs.

-----

I'm not sure how scala-native would help here, as we still need to be running all of our JVM frameworks on the JVM.  This includes Java & Clojure, but also Groovy, Scala, and Kotlin moving forward.

-----

The other option is trying to roll our own, but I think we'll basically end up with something similar to GroovyServ.

kazk:
Added Scala Native because I thought it might be interesting to also consider separating.
If "X Native" existed for all the language supported, this problem can be solved by having `{java,clojure,scala}-runner` instead of `jvm-runner`. I knew about Scala Native and haven't looked into other languages.

[...]

PR #404, opened 2017-04-28
Groovyserv POC

pibi:
related to #326


docker/java.docker
```diff
@@ -0,0 +1,69 @@
...
+RUN groovyserver -t 20 --debug || echo "server connection error"; groovyclient -Ct 120 -Cdebug -e "println 'Hello from GroovyServ'"
+
+# Run the test suite to make sure this thing works
+USER codewarrior
+
+# Set environment variables
+ENV TIMEOUT 120000
+ENV USER codewarrior
+ENV HOME /home/codewarrior
+RUN groovyserver -t 20 --debug || echo "server connection error"; mocha -t ${TIMEOUT} test/runners/java_spec.js
```
jhoffner: I'm curious how this works. It looks like you are running the groovy server within the image as way of prewarming the server within the image itself, but the image doesn't persist application run state, just disk state right?

docker/java.docker
```diff
@@ -0,0 +1,69 @@
...
+
+USER root
+RUN echo '#!/bin/bash\ngroovyserver -t 10 -q || echo ""\n timeout 15 node "$@"'>/runner/entrypoint.sh ; chmod +x /runner/entrypoint.sh
```
jhoffner: Did entrypoint.sh not get committed?

docker/java.docker
pibi: Yep, the image persist only the disk state, not process. the groovyclinet called within node cna start the groovyserver by itself, but it is actually slower than a prewarm.

jhoffner: NVM I think the first `groovyserver` call is you testing that it works and the this call referenced here is just for running the tests. Got it.

docker/java.docker
pibi: it is created within the line 63

pibi: It's created at line 63

jhoffner: oh yeah. duh.

docker/java.docker
```diff
@@ -0,0 +1,69 @@
...
+ENV HOME /home/codewarrior
+RUN groovyserver -t 20 --debug || echo "server connection error"; mocha -t ${TIMEOUT} test/runners/java_spec.js
+
+USER root
+RUN echo '#!/bin/bash\ngroovyserver -t 10 -q || echo ""\n timeout 15 node "$@"'>/runner/entrypoint.sh ; chmod +x /runner/entrypoint.sh 
+
+USER codewarrior
+#timeout is a fallback in case an error with node
+#prevents it from exiting properly
+ENTRYPOINT ["/runner/entrypoint.sh"]
```
jhoffner: OK so still trying to wrap my head around how this will work. You are saying there is a 10 second warmup time that groovyserve costs on initial startup. During that 10 seconds we can't send commands because it will fail, so this delays the call until after 10 seconds? 

That makes sense. In production we pre-warm container before making them available by running `listen.js`, then connect to the container and use `docker exec` to run the actual `run-json.js` file, so everything should be warm by then. 

Have you tested performance by starting using listen.js and then connecting via `docker exec`? I'm curious to know how this performs compared to jvm-runner.

pibi: there's a 10 seconds timeout for the groovyserver startup. Startup time is actully pretty in line with my local setup, that is quite dated. Newer hardware should be faster. 
I'm not sure how you are actually sends the commands, I thought you were using the run.js and a proprietary listen.js. From what you say, I'm wondering if the listen.js is just a way to keep the container up. So, are you sending command with docker exec then?

pibi: By the way, the groovyclient used within the java runner starts the groovyserver automatically if it is not ready, so, during the prewarm you can  submit commands, they will be solved a little bit slower than the others

jhoffner: > From what you say, I'm wondering if the listen.js is just a way to keep the container up. So, are you sending command with docker exec then?

Exactly. `listen.js` is probably a misnomer. It should be `keep-alive.js`. Our scheduling engine will use `docker exec` to a pre-started instance and then kill the instance after it is done, then waiting till the next instance is started before taking on any more work for that node (cluster node, not node JS)

pibi: here is a quick test with docker exec:

$ time docker exec runner-jvm node run -l java -c $'class Solution {\npublic static void main(String[] args){\nSystem.o ut.println("Hello world");\n}}'
Hello world

real    0m3.119s
user    0m0.015s
sys     0m0.031s

jhoffner:
@pibi can you resubmit this PR to the develop branch? I'm about to go on vacation and will test this on a separate branch before I go, but don't want to roll out anything major before I leave.

pibi:
@jhoffner  PR for develop submitted. Unfortunately it looks like it merged the develop branch here too.

Issue #395, opened 2017-04-24
[Proposal] Add Official BF support (as independent language)


DonaldKellett:
*Note: BF is a common acronym for* [Brainf**k](https://esolangs.org/wiki/Brainfuck)

**I know this sounds wild**, but ... 

### Current Situation

Currently, it is entirely possible to write a Kata where the solver has to write a suitable program in BF to pass the tests specified as long as the Kata author knows how to properly implement a bug-free BF interpreter to interpret the user's solution.  However, such Kata are currently counted as a completion in another language entirely (e.g. JavaScript, see [this Kata](https://www.codewars.com/kata/582c2dffd82a1ff03000000a/discuss) and a few others) despite the fact that the user actually needs to know how to program in BF *instead of* that other language that the Kata is written in in order to complete it.  This poses a few problems; for example, if said BF kata is currently available in JavaScript and I translate it into PHP (and it gets approved), did I really translate that Kata into PHP?  Or did I just merely copy the Kata using PHP syntax instead of JavaScript syntax?  And for users who have completed said Kata in both JavaScript and PHP, do they **actually** know either of those two languages, or do they just know BF and instead got free rank progress in both JavaScript and PHP?

### Rationale

The official support of BF as an independent language solves the logical dilemma above and gives the correct credit to the solver of the Kata as having completed a Kata in BF and not in those other languages.

### How To Support BF

Initially, perhaps we can add basic BF support in Kumite Only without a testing framework which I guess should be easy enough as it shouldn't require much more than implementing a working interpreter for it.  Then, in the (hopefully near) future, we could perhaps create a custom testing framework **which is written in another (popular) language such as JavaScript or Ruby** (like we did for SQL) designed to test BF code specifically.  I know that BF is technically "Turing complete" so it is *technically* possible to write tests and preloaded code entirely in BF **but** we don't want to scare people off by making tests insanely hard to write :P

### Additional Notes

Currently, Kata involving writing BF programs do not receive much attention at all from the Community (not a single one of the BF kata authored by [gkucmierz](http://codewars.com/users/gkucmierz/authored) have even come close to approval yet) but I think the problem is **not** because BF is unpopular, but rather because it is not officially credited by the Codewars system as the solver making rank progress in BF which in turn provides no motivation for BF enthusiasts to actually complete such Kata.

Voileexperiments:
[...]

PR #405, opened 2017-04-29
Add Complete BF Support

DonaldKellett:
@jhoffner ,

This pull request aims to add the following features:

- [x] Basic Execution Support of BF on Codewars
- [x] Test Integration (using existing JavaScript CW-2 testing framework) to allow for BF Kata
  - [x] Update `test/runners/bf_spec.js` to include suitable tests for `testIntegration`
  - [x] Edit the `testIntegration` section of `lib/runners/bf.js` to include JavaScript CW-2 framework as testing option such that the newly added tests in the BF Spec pass without any issues
  - [x] Update `.travis.yml` to ensure that Travis tests the `esolangs` image
- [x] Further Improvements
  - [x] Add tests for `<FAILED::>` cases
  - [x] Add tests for `<ERROR::>` cases
  - [x] Add `examples/bf.yml` and test them
  - [x] Add documentation
    - [x] Mention size of memory tape (`3e4` cells as per standard implementation)
    - [x] Mention cell wrapping behavior as per standard implementation
    - [x] Make special note of EOF handling - interpreter currently inputs `-1` to cell under pointer when this happens
    - [x] Give credit to http://www.hevanet.com/cristofd/brainfuck/epistle.html and refer to it whenever necessary
  - [x] Reconfigure number of cells in memory tape from `9999` to `3e4` to comply with standard implementation

All related aspects of basic BF execution have been successfully tested in `test/runners/bf_spec.js` to confirm that it works.  As of 02/05/17, I have yet to add test integration to BF.

In terms of the large number of commits, I apologize for making this pull request look rather messy.  Since this is the first time I’ve actually tried to seriously use the Command Line, Git and Docker to add support to an entire language, I’ve hit quite a few roadblocks in the process which lead me to commit certain changes, only to make another commit to undo it.  I hope you don’t mind and will be able to merge this PR.

Cheers,
DonaldKellett (donaldsebleung on Codewars)

Special credits to @kazk for providing me all the help and support required throughout making this pull request; all this would not be possible without him/her :smile:


kazk:
(reposting my [comment](https://github.com/Codewars/codewars-runner-cli/issues/395#issuecomment-296881515))

@jhoffner how about `language: esolang` and `languageVersion: bf`?
This will enable what @Voileexperiments commented in #395. Tests can be done similar to shell.

kazk:
@jhoffner Looks like GitHub has an option to squash merge.

image

<https://github.com/blog/2141-squash-your-commits>
<https://help.github.com/articles/about-pull-request-merges/>

DonaldKellett:
Bump.  @jhoffner may you please review this pull request?  Thanks a lot :smile:

jhoffner:
@DonaldKellett the PR looks good, however one thing to note is that without `testIntegration` mode we won't be able to use this with kata, only kumite. 

Since Node is already loaded on the machine and BF tends to get implemented within JS kata anyway, it might make the most sense to just use cw-2.js or mocha as the test framework. The other option is to use Ruby to keep it inline with how we did SQL/Bash, though in that case you would have to install BF on the ruby image. Using JS seems like it makes the most sense to me.

> how about language: esolang and languageVersion: bf?

I would probably just make bf its own language, since I'm assuming other eso languages would be different enough to warrant their own translation? IDK I'm not familiar with any of these to really be sure.

kazk:
@DonaldKellett `RECENT_CONTAINERS` is no longer used for Travis, so you need to edit `.travis.yml`.

---

@jhoffner 

> I would probably just make bf its own language, since I'm assuming other eso languages would be different enough to warrant their own translation?

I agree, but what if people wanted to keep adding esolangs?

jhoffner:
> but what if people wanted to keep adding esolangs?

Not sure I follow. There is nothing stopping them from using the same image for multiple languages, possibly each with their own versions. JS, CoffeeScript and Typescript all share the same image but have their own language, with JS having multiple language versions.

kazk:
I just thought it was better if we somehow group them together.

I was concerned that Codewars can get cluttered if we keep adding esolangs as independent languages. There are many esolangs ([esolangs.org](https://esolangs.org/wiki/Language_list) lists 1k+) and it's really easy to add support for them.

kazk:
Just to be clear, I just thought it'll be more flexible for the future requests and not thinking that people will actually open a PR for every esolang they like.

jhoffner:
> There are many esolangs (esolangs.org lists 1k+) and it's really easy to add support for them.

Oh god I hope not. BF is cool, it's the most common one from what I can tell. I see it on a lot of other sites and I can how it would be a good value to have it be an official language. However I think we should be very selective about adding additional ones. For one, support doesn't end with a CLI PR. It has to be documented and added into CW, then supported. Syntax highlighting needs to be updated, additional CodeMirror languages may need to be installed. Then people start asking for version issues, or someone finds some online converter that makes it easy to just add crappy translations (like CoffeeScript) and content quality ends up being an issue...

kazk:
I agree. If we're being selective, then I don't think BF being independent language will be a problem.

I wrote "easy to add" because

- most of eslolangs can be added to run by simpy doing `[esolang] [options] file [optional-inputs]`
- no need to *choose* the test framework in that language because it doesn't exist
- outputs can be tested using already supported test frameworks


> For one, support doesn't end with a CLI PR. [...] Syntax highlighting needs to be updated, additional CodeMirror languages may need to be installed.

I recently learned that finding usable language icon set (with uniform style) is really hard...

DonaldKellett:
(Note to self)

TODO:

- Edit the `testIntegration` section of `lib/runners/bf.js` to include JavaScript CW-2 framework as testing option
- Update `test/runners/bf_spec.js` to confirm that `testIntegration` works
- Update `.travis.yml` to ensure that Travis tests the `esolangs` image

(*Feel free to ping me though in case I have missed anything from this list, thanks a lot* :smile:)

DonaldKellett:
@jhoffner A question regarding how BF is currently added in this pull request:

From what I understand, the current setup is such that if it is implemented on Codewars, the user will write their BF program in the "Solution" window, then when they click "Run" (or "Publish" when they publish the Kumite), the BF program is immediately executed *and only once* and its STDOUT printed to the console window.  With regards to this, I was wondering how we would be able to test a BF program with different inputs (which is required for random testing and preventing "hardcoded" BF programs) when test integration is added?  Or should I modify it such that the testing framework (JavaScript CW-2 or otherwise) will be able to access the source code of the user solution and execute it at will using a custom BF interpreter implemented in JavaScript?

jhoffner:
> Or should I modify it such that the testing framework (JavaScript CW-2 or otherwise) will be able to access the source code of the user solution and execute it

Yes this is the approach I would take. You can look at the SQL and Shell implementations for how we did those in Ruby. You could either use Ruby (you would need to add to that image instead) or you could do it instead in JavaScript (cw-2, mocha). 

The test integration would mostly be you building a little API around calling the bf code. Something along the lines of:
```javascript
const execSync = require('child_process').execSync;
module.exports.runBf = function (input) {
  return execSync(`bf /workspace/solution.txt`, { input })
}
```

Its similar to what you did for solutionOnly mode, except that you would be providing this functionality for running within a node/mocha executable. 

So you would execute a new node/mocha process using `runCode` and then within that process you would execute a `bf` process.

I'm pretty tired so take this for whatever it's worth. Hopefully this makes sense. It's basically inception (the movie).

DonaldKellett:
Many thanks for your reply and help :smile:  I think I'll stick with JavaScript since I am more familiar with it.

DonaldKellett:
@jhoffner Thanks for the code snippet and I have included it in my `testIntegration` but I still fail to see how the code allows me to execute BF in conjunction with the current existing JavaScript CW-2 framework.  For example, when I do the testing, do I do something like the example below (in `test/runners/bf_spec.js`) to confirm that it works?

```javascript
Test.assertEquals(runBF('some input that is not needed for this particular input-independent BF program'), "Hello World!");
```

@kazk May you provide some hints on this one if possible?  I've taken a look at the `sql.js` and `shell.js` and I see that they utilize `ruby.js` somewhere in their code but I still fail to see how it works by simply passing its arguments to `(ruby.)prepareCW2` when the code involved is obviously written in another language.

kazk:
@DonaldKellett I don't have much time right now, so just some quick notes.

To see how `shell` is tested, see [frameworks/ruby/shell.rb](https://github.com/Codewars/codewars-runner-cli/blob/815480438a336443327e850a2c9e51eaae8e4989/frameworks/ruby/shell.rb).

---

For handling inputs:

I looked at example usage of `bf` and it seems to take input from `stdin`.

```
bf cat.b < file
```

Maybe [`beef`](http://packages.ubuntu.com/trusty/beef) is easier to understand for you because it allows to take a file as input. See [`beef(1)`](http://manpages.ubuntu.com/manpages/trusty/man1/beef.1.html).

```
beef --input-file=input code.b
```

---

Implement `runBF([input])` like @jhoffner described.

```javascript
execSync(`bf /workspace/solution.txt`, { input });
// or
execFileSync('bf', ['/workspace/solution.txt'], { input });
// or
// write input to file `input.txt` then
execFileSync('beef', ['--input-file', 'input.txt', '/workspace/solution.txt']);
```

Including that function in fixture will allow:

```javascript
// for no input case, don't do `{ input }`
Test.assertEquals(runBF(), expected);
Test.assertEquals(runBF(input), expected);
```

---

I realized that this comment isn't adding much information to @jhoffner's. I think I misunderstood what you're asking.
Can you describe your issues more specifically? I'll try to help again when I have some time.

DonaldKellett:
> Including that function in fixture will allow:
> 
> ```javascript
> // for no input case, don't do `{ input }`
> Test.assertEquals(runBF(), expected);
> Test.assertEquals(runBF(input), expected);
> ```

Thanks, that's all I need to know *for now* - at least with these tests written in `bf_spec.js` I'll know whether the runner is running as expected or not :smile:

DonaldKellett:
@kazk I've noticed that the `testIntegration` for both the SQL and Shell runners first do a bit of setup and then just pass in the correct arguments to `ruby.prepare[TEST_FRAMEWORK_NAME]` where `ruby === require('./ruby')`.  So I tried to copy this into my BF runner by doing `var javascript = require('./javascript')` and then did the same, exporting `module.exports.runBF` as per your instructions (and @jhoffner's) and then calling on `javascript.prepareCw2` but then I receive a `TypeError` saying that that isn't a function.  Any idea why that might occur?  I've checked the JavaScript runner and there indeed exists a function called `prepareCw2` which is pretty well defined.

jhoffner:
BTW @DonaldKellett I know you are going to be excited to get this deployed to Codewars once this is ready. FYI I will be going on vacation starting May 12th and will be gone for a month, with the first 2 weeks me being fully on vacation and the next 2 weeks after I will still be in Bali but will be getting some work done. 

Up to and during this time I will not be pushing any major changes, to make sure that stability holds well while I am away. Thanks for your contributions and once this is ready I will get it out as soon as possible once I'm back.

kazk:
@DonaldKellett

> calling on javascript.prepareCw2 but then I receive a TypeError saying that that isn't a function. Any idea why that might occur?

This is because `javascript.js` doesn't export `prepareCw2`. It also does more than necessary for this. Executing something like the following should be enough. 


```javascript
const fixture = `
require('/runner/frameworks/javascript/cw-2');
const runBF = require('/path/to/run-bf.js');
${opts.fixture};
`;
```

---

If you're doing `module.exports.runBF` you need to put that in a separate file and do `const runBF = require(path to the file)` (see above).
You can also just prepend `runBF` to `opts.fixture`.

---

I wanted to help you to understand *this codebase*, but I'm starting to feel it's getting out of scope. If you're not familiar with how Node's modules work, then I'd recommend learning that first.

DonaldKellett:
@jhoffner No worries, have a nice vacation :smile:  In case I do manage to get this finished without any bugs and such at least 2 days before your vacation starts, will you be able to add BF on Codewars beforehand?

> I wanted to help you to understand this codebase, but I'm starting to feel it's getting out of scope. If you're not familiar with how Node's modules work, then I'd recommend learning that first.

Thanks, I'll consider learning Node.js seriously after I get this pull request done and merged (I really want to get this in well before May 12 if possible).  I'll have to say though, I've tried to learn Node before using nodeschool.io (or is it `.com`?) with little success (I think I got stuck on async functions and that) but hey, that was like well over a year ago (before I even joined Codewars) so I'll take another shot :smile:  Oh, and thanks for your code snippet again, I'll work on it today in my spare time.

DonaldKellett:
@kazk Sorry to bother you yet once again but I'm struggling quite badly in the BF Test Integration.  As per your instructions, I have added a `run-bf.js` file in `frameworks/bf/` (I wasn't quite sure where to put it but I thought that was the most sensible place) with the following code:


```javascript
var execSync = require('child_process').execSync;
module.exports.runBf = function(input) {
  if (typeof input != "undefined") {
    return execSync('bf /workspace/solution.txt', {input});
  }
  return execSync('bf /workspace/solution.txt');
};
```

And my `testIntegration` is as follows:

```javascript
function(runCode, fail) {
  switch (opts.testFramework) {
    case 'cw-2':
      var fixture = '\
require(\'/runner/frameworks/javascript/cw-2\');\
const runBF = require(\'/runner/frameworks/bf/run-bf.js\');\
' + opts.fixture + ';\
';
      runCode({
        name: '/usr/local/n/versions/node/6.6.0/bin/node',
        args: ['-e', fixture, '--no-deprecation'],
        options: {
          cwd: opts.dir
        }
      });
      break;
    default:
      throw 'Test framework is not supported';
  }
}
```

But then, when I run `make esolangs`, I get the following error when the `bf_spec.js` assertions reach the Test Integration section:

```
Error: spawn /usr/local/n/versions/node/6.6.0/bin/node ENOENT
```

... and the STDOUT just ends up as an empty string `''`.  May you please explain to me what this error means?  Thanks a lot :smile:

kazk:
You're getting that because `/usr/local/n/versions/node/6.6.0/bin/node` doesn't exist.
`n` is a Node version manager and it's used in `node-runner` to support multiple Node versions (`0.10` and `6`).

Try replacing that with just `node` which should use the first `node` found.

DonaldKellett:
Thanks @kazk for the quick reply, I replaced all that with just `node` and the errors indeed disappeared but for some reason the STDOUT is still just an empty string.  Could that be because I've somehow passed in an empty file to the `bf` command in my `runBF`?  I don't seem to remember having written anything to `/workspace/solution.txt`.

kazk:
`/workspace/solution.txt` is `/home/codewarrior/solution.txt` and should contain `opts.solution`. I think it's done inside `lib/shovel.js`.

DonaldKellett:
@kazk So does that mean I should include `opts.solution` somewhere in my call to `runCode` in the test integration?

frameworks/bf/run-bf.js
```diff
@@ -0,0 +1,7 @@
+var execSync = require('child_process').execSync;
+module.exports.runBf = function(input) {
```
kazk: If you have `module.exports.runBf`, `require('/runner/frameworks/bf/run-bf.js');` will be `{runBf}`.

Try doing `module.exports = function runBF(input)` instead.

You can also do `const runBF = require('/runner/frameworks/bf/run-bf.js').runBf;` but I think it's unnatural.

Another option is just embed it directly after

```js
require('/runner/frameworks/javascript/cw-2');
```

which might make more sense because it's just a single function.

DonaldKellett: Thanks, I'm now finally getting a non-empty STDOUT.  It appears that the `bf` fast Brainf**k Interpreter doesn't just output the expected string though - it outputs a buffer instead with hexadecimals - but that shouldn't be too hard to fix, I think (maybe `5/6 kyu` in Codewars terms :wink:)

kazk: Try adding option `encoding: 'utf8'` for `execSync`, or just call `.toString()` on returned buffer.

DonaldKellett: Actually @kazk to be specific, it appears that I'm now getting a string of the format `<Buffer 48 65 6c 6c 6f 20 57 6f 72 6c 64 21>` (which is supposed to be "Hello World" in this example) when `runBF` is executed.  I've tried to use a RegExp replace on it, trying to treat the return value of `execSync` as a plain string but somehow the entire STDOUT becomes empty again instead.  I'm pretty sure that my RegExp is correct (after all, I'm just trying to remove the `<Buffer ` and `>` parts), but may you please help me check out this (commented out) part of my `runBF` function?

```javascript
return execSync('bf /workspace/solution.txt', {input})/* .replace(/^\<Buffer |\>$/g, "") */;
```

Or is my treatment of the return value of `execSync` as a plain string incorrect?

kazk: `execSync` returns `Buffer` by default and that's a string representation of `Buffer`. Nothing to do with `bf`.

```js
console.log(new Buffer("Hello World!"))
//=> <Buffer 48 65 6c 6c 6f 20 57 6f 72 6c 64 21>

console.log(new Buffer("Hello World!").toString('utf8'))
console.log(new Buffer("Hello World!").toString())
//=> Hello World!
```

kazk:
```js
const runBF = input => execSync('bf /workspace/solution.txt', input != null ? {input, encoding: 'utf8'} : {encoding: 'utf8'})
// or
const runBF = input => execSync('bf /workspace/solution.txt', input != null ? {input} : {}).toString()
```

DonaldKellett:
@kazk I'm now getting very close to get this thing actually running (many thanks to you :smile: :tada:) but it appears that the conversion of Buffer to String doesn't work properly whenever the extended ASCII character set is included.  For example, for my BF Multiplication program and the input `String.fromCharCode(15, 12)`, the output is supposed to be `String.fromCharCode(180)` (which I just learned is the backtick from the failed test output) but instead it gives this: `�`.  I know some languages (e.g. PHP) that cannot properly handle the extended ASCII but I've never seen this happen with JavaScript.  Do you know how this problem could be fixed, or should I just avoid such tests involving extended ASCII altogether?

kazk:
@DonaldKellett I don't think there's any need to add dependencies. That question is about *reading* a file not in utf8.

>> The file I was trying to read was in ANSI encoding. When I tried to read it using the 'fs' module's functions, it couldn't perform the conversion of the extended ASCII characters.
>>
>> I just figured out that nodepad++ is able to actually convert from some formats to UTF-8, instead of just flagging the file with UTF-8 encoding.
>>
>> After converting it, I was able to read it just fine and apply all the operations I needed to the content.
>
> Thank you for your answers!

---

Try passing `input` as `Buffer`.

DonaldKellett:
> Try passing `input` as `Buffer`

So something like `execSync('bf /workspace/solution.txt', {new Buffer(input)})` in my `runBF` function?

kazk:
I might have misunderstood your issue. That SO was about reading (input) and your issue is output?

DonaldKellett:
@kazk Yes, my issue is output.  Although I haven't tested extended ASCII *input* yet in my `bf_spec.js` assertions - should I add those as well?

kazk:
Try

```js
const runBF = input => execSync('bf /workspace/solution.txt',
  input ? {input, encoding: 'utf8'} : {encoding: 'utf8'})
```

The documentation says "The encoding used for all stdio inputs and outputs. "

DonaldKellett:
Thanks, but the result is still the same :(  Guess we'll have to wait for @jhoffner to add the dependency so I can use it for my `runBF`.

kazk:
I don't think that'll solve the issue though. Why don't you just try it locally?

Can you post the entire output?

kazk:
@DonaldKellett is this what you mean?

```js
const b = Buffer.alloc(1);
b[0] = 180;
console.log(b.toString('utf8'), String.fromCharCode(...b));
� ´
```

DonaldKellett:
Yeah, I think that's it :smile:  So does that mean if I do `String.fromCharCode(...b)` where `b` is the currently incorrectly encoded string, the problem will solve itself?

kazk:
`b` is a `Buffer`. To be honest, I'm not really sure why this works. You can try it.

DonaldKellett:
Thanks @kazk I just tried it and it worked :tada:  Another task off the list! :smile:

DonaldKellett:
And ... done!  Let's hope that the Travis build passes without any issues :smile:

DonaldKellett:
@jhoffner The most recent Travis build passed for all images except for Julia (which has nothing to do with the changes proposed in this pull request); please merge this pull request if satisfactory :smile:

kazk:
I just restarted the job.

kazk:
<https://travis-ci.org/Codewars/codewars-runner-cli/jobs/228661257#L579>
```
gpg: requesting key 03D6E495 from hkp server ha.pool.sks-keyservers.net
gpg: keyserver timed out
gpg: keyserver receive failed: keyserver error
```

docker/esolangs.docker
```diff
@@ -0,0 +1,34 @@
...
+FROM codewars/base-runner
+
+# Install "bf" Fast Brainf**k Interpreter
+RUN apt-get update && apt-get install -y --no-install-recommends bf
+
+# Install "beef" flexible Brainf**k Interpreter (just in case)
+RUN apt-get update && apt-get install -y --no-install-recommends beef
+
+# Install "bfs-buffer" to deal with extended ASCII characters
+RUN npm install bfs-buffer
```
kazk: Remove `beef` and `bfs-buffer` if it's not used.

lib/runners/bf.js
```diff
@@ -0,0 +1,34 @@
...
+      });
+    },
+    testIntegration: function(runCode, fail) {
+      switch (opts.testFramework) {
+        case 'cw-2':
+          var fixture = '\
+require(\'/runner/frameworks/javascript/cw-2\');\
+const runBF = require(\'/runner/frameworks/bf/run-bf.js\');\
+' + opts.fixture + ';\
+';
```
kazk: I don't think this codebase uses `\` anywhere (at least for JavaScript).
I'd suggest changing them to keep consistency.

docker/esolangs.docker
DonaldKellett: Done :smile:

lib/runners/bf.js
DonaldKellett: Done :smile:

lib/runners/bf.js
```diff
@@ -0,0 +1,34 @@
...
+        args: [solutionFile]
+      });
+    },
+    testIntegration: function(runCode, fail) {
+      switch (opts.testFramework) {
+        case 'cw-2':
+          var fixture = `
+require(\'/runner/frameworks/javascript/cw-2\');
+const runBF = require(\'/runner/frameworks/bf/run-bf.js\');
+${opts.fixture};
```
kazk: All of those `\'` should be unnecessary.

test/runners/bf_spec.js
```diff
@@ -0,0 +1,96 @@
...
+. Print \"l\"\
++++. Print \"o\"\
+<<<<++. Print \" \" (spacebar character)\
+>>---. Print \"W\"\
+>>. Print \"o\"\
++++. Print \"r\"\
+------. Print \"l\"\
+<-. Print \"d\"\
+<<<+. Print \"!\""}, function(buffer) {
+        expect(buffer.stdout).to.equal("Hello World!");
```
kazk: `\`

there's more below

lib/runners/bf.js
DonaldKellett: Oops, sorry, I forgot to update those, my bad 😛 Those unnecessary backslashes should be removed now :smile:

test/runners/bf_spec.js
DonaldKellett: And done :smile:

DonaldKellett:
> What happens when `bf` gets invalid code?
`execSync` will throw if `bf` has `non-zero` exit code.
This needs to be looked into and tested.

@kazk Invalid code as in unmatched square brackets `[]`?  Sure, I can implement a square bracket check on the code located in `/workspace/solution.txt` before I pass it on to `execSync`, but what happens if the code turns out to be invalid?  Should my function throw an `Error`?

kazk:
I don't know how `bf` handles them, but `exitSync` will throw if `bf` gives non-zero exit code.
The thrown error should be caught by the test framework, but the error message might not be informative at all.

Write some tests with invalid codes and see how things are handled. See if `bf` can show any useful info. Try to see the output in the solver's point of view and improve the output.

DonaldKellett:
@kazk I've just added 4 tests in `bf_spec.js` testing specifically for the STDERR output caused by invalid BF syntax (pointer out of bounds, unmatched brackets, etc.).  So far, the errors I've seen thrown directly by the `bf` command have been quite informative, providing a specific reason as to why the error occurred.  Sometimes, however, the `bf` fast Brainf**k Interpreter deals with unmatched brackets as a pointer out-of-bounds error instead, but that could just be due to the way that it iterates through the BF code.  After the build for the most recent commit is complete, if you read the Travis output and think that the error messages thrown are not informative enough (or inaccurate), feel free to inform me and I'll edit the `runBF` function such that it throws errors with more informative error messages.  Otherwise, feel free to inform me if I still need to **add** anything (files, major features, etc.) :smile:

@jhoffner If the most recent Travis CI build succeeds for all images except for Julia like the last 2 or so commits, please consider merging this pull request even if the actual changes won't be implemented on Codewars until June, thanks a lot :smile:

kazk:
- Add tests for `<FAILED::>` cases
- Add tests for `<ERROR::>` cases
- Add `examples/bf.yml` and test them
- Add documentation
- Decide number of cells allowed (`-c`, defaults to `9999`)
- Decide what to do with wraparound (`-w`, defaults to allow) and document that

DonaldKellett:
@kazk Through more rigorous testing of the behavior of the `bf` command, I've discovered that this particular interpreter stores a value of `-1` to the cell under the pointer when EOF has been reached in the input stream.  *Although* this behavior is documented as a possible standard implementation for handling EOF, more natural behaviors would include either storing `0` instead or leaving that cell unchanged.  Do you know if there are any flags I can set for this command that changes its behavior upon reaching EOF?  If not, that's fine too, since the current behavior is considered "standard" to an extent, I can just document it and make it aware to BF solvers.

DonaldKellett:
> Add `examples/bf.yml` and test them

Just to clarify, but these two examples don't have to be "Two Oldest Ages" and pushing `"codewars"` to an empty array (however that's understood in the context of BF 😛), right?  I'm not an expert in writing BF programs and will seriously struggle with "Two Oldest Ages".

kazk:
@DonaldKellett 

For `bf` flags, see <http://manpages.ubuntu.com/manpages/trusty/man1/bf.1.html>.

I think you can be creative with examples because BF is very different from other languages.
For BF, small "bug fixes" can be hard :) "refactoring" can be making the given code shorter/efficient.

DonaldKellett:
Thanks for the link to the manual, it appears that there is no such option to alter the behavior of the interpreter when reaching EOF.

> I think you can be creative with examples because BF is very different from other languages.
For BF, small "bug fixes" can be hard :) "refactoring" can be making the given code shorter/efficient.

I guessed so, thanks for the confirmation :smile:

DonaldKellett:
Done @kazk @jhoffner, please review 😄

kazk:
@DonaldKellett 

Great job with the documentation :+1:

For the tests "should provide a useful error message for invalid BF code", messages are from `bf` and just written to `stderr`? It's hard to see from the Travis output but the test case is `expect(buffer.stderr).to.contain('Error')`.

I was thinking that in case of invalid code, `bf` will have non-zero exit code causing `execSync` to throw. This seems to happen in "should handle errors properly when spec methods are being used" test.

@jhoffner, this looks like a test framework bug to me.
If the test case is inside `describe/it`, `Test.handleError` is used. If not, it's handled by `uncaughtException` on `process` which just ignores if the test is synchronous.

```javascript
process.on('uncaughtException', function(err) {
  if (async) {
    Test.handleError(err);
    if (asyncDone) asyncDone();
  }
});
```

jhoffner:
@DonaldKellett first glance at this it looks really good. Great job! I need to run out for a bit but when I get back I'll dig in deeper. Thanks @kazk for all of the support in this.

One thing to note, I wonder if this should just be installed on the node image. @kazk otherwise once we start splitting out the runners based off of your new modular design, we will need to share cw-2.js somewhere shared.

DonaldKellett:
Thanks for the compliments :smile:

> For the tests "should provide a useful error message for invalid BF code", messages are from `bf` and just written to `stderr`? It's hard to see from the Travis output but the test case is `expect(buffer.stderr).to.contain('Error')`.

@kazk I thought that would have been suitable because the error messages provided by `bf` are pretty comprehensive IMO, but should I do something like a `try/catch` block instead and throw my own error?  Or should `runBF` swallow the error and just `console.log` a message containing `<ERROR::>` instead?

> One thing to note, I wonder if this should just be installed on the node image.

@jhoffner Should I proceed with this now or should I wait until a decision is reached?

kazk:
@DonaldKellett 

> should I do something like a `try/catch` block instead and throw my own error? Or should runBF swallow the error and just `console.log` a message containing `<ERROR::>` instead?

No, I think it's test framework's bug because the thrown error is ignored. You don't need to do anything until it's confirmed and fixed. Let's wait for @jhoffner's review.

For using `node-runner` or not, I'd say we keep this as is for now. There're still some things that I'd like to consider and discuss before deciding.

DonaldKellett:
Bump.

@jhoffner @kazk Has a decision been reached yet regarding whether the bf-related installations should just be installed on the Node image instead?

jhoffner:
I'm fine either way. If we split the repos we can deal with how to organize things then.

kazk:
Just keep them separate for now. I don't think it's worth rushing the decision and this shouldn't affect the code added in this PR much.

According to my comments (https://github.com/Codewars/codewars-runner-cli/pull/405#issuecomment-299654449, https://github.com/Codewars/codewars-runner-cli/pull/405#issuecomment-299681924), there's a possible bug in the test framework waiting for @jhoffner's input that might affect tests included in PR.

---

@DonaldKellett
Please understand that reviewing a PR is a lot of work, especially when it's not organized.
To be honest, I'm kind of lost now. I have to re-read some comments to refresh.
Next time, try your best to make logical commits to make everyone's life easier.
Maybe I gave the wrong impression by commenting about GitHub's squash merge button at the beginning of this PR. My intention was for @jhoffner to use it to keep the history sane without him doing extra work and not to encourage this style.

I consider this project too relaxed :sweat_smile:
Many projects won't even look at your code if PR looked like this.
I apologize if I sounded little harsh and I don't mean to discourage your contributions at all.
I wanted you to realize that better commits will result in faster/more feedbacks.
It's really hard to motivate yourself to review a PR with 40+ commits for 8 files having 70+ comments.
I'd close my browser if I wasn't already involved :laughing:

DonaldKellett:
My most recent comment was just to ensure that this PR hasn't been completely forgotten 😅; after all, I've spent quite a few hours making this PR at the expense of another client project I'm working on (and my much-needed revision for my A-Level exams which I'm really focusing on now - that's why I haven't touched a single line of code for about a month now) and would be a little disappointed if it was completely dismissed after a month or so.

****

@kazk Apologies for the huge mess I've made in making this PR - as you are aware, that was (and still is) my first and only time working with Docker so I naturally ran into a lot of issues which caused me to make (and undo) a lot of small changes in the process.  However, I do have to admit that I made commits *way* more often than necessary just to make sure that even the tiniest changes I made in my local copy of the repo are backed up on GitHub (just a really bad habit of mine - sorry).  Next time I make a PR (on this Repo or otherwise) I'll make sure that I only commit changes that are significant to the project such as writing a full, functioning file or adding a feature that has already been rigorously tested and proven locally.

Now that I look back at this, I do understand your (and quite possibly @jhoffner 's) pain in re-reviewing the entire pull request, seeing as even the comment section itself might take the average person a full hour to read, if not more.  I really do appreciate all the time, effort and support you have put into this pull request to make it into a reality and would be absolutely flattered if this pull request was reviewed and approved within the next month or two.

kazk:
@DonaldKellett
I made #434 from this PR and included relevant comments. Let me know if I missed anything.
If everything is fine, let's continue this on #434.

DonaldKellett:
> By the way, please look into `Test.assertApproxEquals` bug (Codewars/codewars.com#962) when you have time.

Sure, I'll look into it once I'm less busy :smile:

> I made #434 from this PR ... If everything is fine, let's continue this on #434.

Thanks a lot for helping me clean up this messy PR; it sure looks a lot neater :smile: :tada:  I just reviewed the changed files and do not believe that you have missed anything so I guess we can now switch to that PR entirely and close this one :smile:

Issue #398, opened 2017-04-24
Server Execution Error "Unexpected token { in JSON at position 1240339"


jhoffner:
@simonvpe commented on [Fri Feb 24 2017](https://github.com/Codewars/codewars.com/issues/843)

I get these errors a lot. Sometimes only on the "Run Examples" build, and sometimes on the "Attempt" build. If i change a line, or add/remove a write to stdout it changes to either not appear at all, or to appear on only one of the builds.

Sometimes I get HTTP 5xx errors or timeouts saying that I write too much to stdout (even though I don't).

This Kata is runs very heavy optimization tests so it's kind of taxing on the servers, is this the issue? Should there be more robust error handling when the server can't cope with the load?

I might add that I can build about one in tweny times so it is very time consuming and ruins the experience completely!

BR Simon

---

@menip commented on [Sun Feb 26 2017](https://github.com/Codewars/codewars.com/issues/843#issuecomment-282574308)

I'm having the same issue with the [longest_palindrome](https://www.codewars.com/kata/54bb6f887e5a80180900046b/train/python) kata. My code:
```python
def longest_palindrome (s):
    longest = 0
    for i in range(len(s)):
        si = s[i:]
        for j in range(len(si)):
            sj = si[:j+1]
            if sj == sj[::-1] and longest < len(sj): 
                    longest = len(sj)
    return longest
```

EDIT: Was able to attempt and submit without issue after an hour passed. 

---

@ioncrash commented on [Wed Mar 08 2017](https://github.com/Codewars/codewars.com/issues/843#issuecomment-285206522)

I've been getting similar errors continuously for about an hour. Position number seems to be different every time I run it

---

@Khanthulhu commented on [Mon Apr 24 2017](https://github.com/Codewars/codewars.com/issues/843#issuecomment-296845184)

Having the same problem with Simple Fun #27: Rectangle Rotation. The position changes each time I submit.

Khanthulhu:
I removed the puts statements from my code and now I don't have a problem. I think it might have to do with the large number of output my program had.

jhoffner:
yeah it's definitely caused by that. This is a bug in that somehow the message telling you that your output is too long is not properly being massaged.

kazk:
@jhoffner
Are you using `opts.format=json` and parsing the entire output as `JSON`?
Something similar happened to me when I was running tests using `run-json`.
[...]

PR #408, opened 2017-05-03
Fix for #398: Ably no longer breaks JSON output

jhoffner:


lib/options.js
```diff
@@ -124,6 +124,9 @@ function assignPublishing(opts) {
 
   if (opts.ably && opts.channel) {
     try {
+      opts.ably = opts.ably === true ? {} : opts.ably || {};
+      opts.ably.log = { level: 0 };
+
```
kazk: This looks unnecessarily complicated to me.

```javascript
opts.ably = opts.ably === true ? {} : opts.ably || {};
//                                    ^ is always truthy because of `if (opts.ably && opts.channel)`
// so
opts.ably = opts.ably === true ? {} : opts.ably;
// or
if (opts.ably === true) opts.ably = {};
opts.ably.log = { level: 0 };
```

What is the current type of `opts.ably`?
I usually avoid changing types like that because it's confusing and triggers de-optimization.
If this is to allow overriding options for Ably through `opts.ably`, how about

```javascript
const ablyOpts = {log: {level: 0}};
if (typeof opts.ably != 'boolean') Object.assign(ablyOpts, opts.ably);
```

jhoffner: Good call on opts.ably already being known to be true and deoptimization. However I wouldn't want to use the assign because no matter what we want the log level to be 0.

Now that I think about it more I'm just not going to make the options configurable at all via the API.

lib/options.js
```diff
@@ -124,10 +124,7 @@ function assignPublishing(opts) {
 
   if (opts.ably && opts.channel) {
     try {
-      opts.ably = opts.ably === true ? {} : opts.ably || {};
-      opts.ably.log = { level: 0 };
-
-      var ably = new require('ably').Rest(opts.ably);
+      var ably = new require('ably').Rest({log: {level: 0}});
```
kazk: I like this much better :)
We can consider making this configurable when it becomes necessary.

Also, I think we should lock the version of Ably in `package.json`. At least the major version.

`"ably": "*"` 😨 

image

`"ably": "^4.x.x"`
image

<https://semver.npmjs.com/>

OverZealous: Definitely make sure to at least set NPM package versions to `^x.y.z`. I'd also **highly** recommend just using [`npm shrinkwrap`.](https://docs.npmjs.com/cli/shrinkwrap).

There's really no reason to allow for different versions across different build environments. Best to lock everything in and make all package upgrades explicit.

kazk:
I think Travis is failing for `push` because `develop` branched from `master` before merging #402 which contained the fix for #400.

kazk:
Looks good to me.
I was surprised to see `ably` `^1.0.4` because I thought `ably` had `4.x.x` versions (the image I posted above). That was the list of `lodash` versions loaded by default and I had to **submit** the package name in order to update the list of versions. Now it looks like this:

image

Issue #410, opened 2017-05-04
Fixing the LF issue


kazk:
Fixing the issue where users' output affects the test result.

```
user logged output<FAILED::>
```

- [x] Dart
  - `Server Execution Error`, see the [comment below](https://github.com/Codewars/codewars-runner-cli/issues/410#issuecomment-299260220)
- [x] C# (#411, #572)
- [x] JavaScript (#502)
- [x] TypeScript (#502)
- [x] OCaml (#468)
- [x] Swift (#451)
- [x] PHP (#421)
- [x] F# (#419)
- [x] Elixir (#417)
- [x] Crystal (#416)
- [x] Ruby (#414)
- [x] C++ (#413)
- [x] Python (#412)
- [x] Lua (#409)
- [x] Java (#336)
- [x] Go (#339)
- [x] C ([@25e5609](https://github.com/Codewars/codewars-runner-cli/commit/25e560940a9cdc249d00e6d732c2a54e0a7014d5))

---

- [x] Rust ([lib/runners/rust.js](https://github.com/kazk/codewars-runner-cli/blob/815480438a336443327e850a2c9e51eaae8e4989/lib/runners/rust.js)) outputs the result differently from the most and not really affected.
  - `stdout` is transformed after the test and only contains the output commands
    ```
    <DESCRIBE::>returns_expected
    <FAILED::>Test Failed
    ```
  - `stderr` contains the details on failure (`print!("foo"); assert_eq!(1, 2);`)
    ```
    Failure Info:
    
    ---- returns_expected stdout ----
    	foothread 'returns_expected' panicked at 'assertion failed: `(left == right)` (left: `1`, right: `2`)', main.rs:17
    ```

---

Codewars/codewars.com#867
Codewars/codewars.com#937

kazk:
C# example from @Voileexperiments's comment

```csharp
using System;
using NUnit.Framework;

[TestFixture]
public class KataTestClass
{
    [Test]
    public void Test()
    {
        Console.Write("foobar");
        Assert.AreEqual(true, false);
    }
}
```

Outputs

```
foobar<FAILED::>  Expected: True
  But was:  False
```

![image](https://cloud.githubusercontent.com/assets/14909989/25662584/a00a9140-3047-11e7-8aa1-16a089f15c56.png)

Voileexperiments:
JS example:
```
Test.expect(true);
process.stdout.write('foobar');
Test.expect(false);
process.stdout.write('foobar\n');
[...]

PR #411, opened 2017-05-04
[C#] Fix LF issue

kazk:
I think this will fix the issue.
I didn't realize that I had to include the compiled binary until I saw the unchanged output on Travis.
Can anyone do that?

---

#410


jhoffner:
Yeah currently I've gotta build the thing locally. Its a terrible setup the whole thing needs to be reworked. .NET's build system is miserable to work with, especially without all of their GUI tools. 

I'll follow up with this tomorrow.

kazk:
Closing this. See #572

PR #434, opened 2017-06-10
Add BF

kazk:
This is just #405 cleaned up.

image

---

kazk commented on [May 6](https://github.com/Codewars/codewars-runner-cli/pull/405#issuecomment-299654449)

> @DonaldKellett 
> 
> Great job with the documentation :+1:
> 
> For the tests "should provide a useful error message for invalid BF code", messages are from `bf` and just written to `stderr`? It's hard to see from the Travis output but the test case is `expect(buffer.stderr).to.contain('Error')`.
> 
> I was thinking that in case of invalid code, `bf` will have non-zero exit code causing `execSync` to throw. This seems to happen in "should handle errors properly when spec methods are being used" test.
> 
> @jhoffner, this looks like a test framework bug to me.
> If the test case is inside `describe/it`, `Test.handleError` is used. If not, it's handled by `uncaughtException` on `process` which just ignores if the test is synchronous.
> 
> ```javascript
> process.on('uncaughtException', function(err) {
>   if (async) {
>     Test.handleError(err);
>     if (asyncDone) asyncDone();
>   }
> });
> ```

jhoffner commented on [May 6](https://github.com/Codewars/codewars-runner-cli/pull/405#issuecomment-299658659)

> @DonaldKellett first glance at this it looks really good. Great job! I need to run out for a bit but when I get back I'll dig in deeper. Thanks @kazk for all of the support in this.
> 
> One thing to note, I wonder if this should just be installed on the node image. @kazk otherwise once we start splitting out the runners based off of your new modular design, we will need to share cw-2.js somewhere shared.

DonaldKellett commented on [May 6](https://github.com/Codewars/codewars-runner-cli/pull/405#issuecomment-299678547)

> Thanks for the compliments 😄
>
>> For the tests "should provide a useful error message for invalid BF code", messages are from `bf` and just written to `stderr`? It's hard to see from the Travis output but the test case is `expect(buffer.stderr).to.contain('Error')`.
>
> @kazk I thought that would have been suitable because the error messages provided by bf are pretty comprehensive IMO, but should I do something like a `try/catch` block instead and throw my own error? Or should `runBF` swallow the error and just console.log a message containing `<ERROR::>` instead?
>
>> One thing to note, I wonder if this should just be installed on the node image.
>
> @jhoffner Should I proceed with this now or should I wait until a decision is reached?

kazk commented on [May 6](https://github.com/Codewars/codewars-runner-cli/pull/405#issuecomment-299681924)

> @DonaldKellett
>
>> should I do something like a `try/catch` block instead and throw my own error? Or should `runBF` swallow the error and just `console.log` a message containing `<ERROR::>` instead?
>
> No, I think it's test framework's bug because the thrown error is ignored. You don't need to do anything until it's confirmed and fixed. Let's wait for @jhoffner's review.
> For using node-runner or not, I'd say we keep this as is for now. There're still some things that I'd like to consider and discuss before deciding.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ env:
     - TEST_IMG=objc
     - TEST_IMG=go
     - TEST_IMG=lua
+    - TEST_IMG=esolangs
 
 script:
   - eslint '**/*.js'
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,12 @@
 HOSTNAME=codewars
 
 # Building erlang images have been suspended (frozen) until they are able to be repaired
-CONTAINERS=node dotnet jvm python ruby alt rust julia systems dart crystal ocaml swift haskell objc go lua
+CONTAINERS=node dotnet jvm python ruby alt rust julia systems dart crystal ocaml swift haskell objc go lua esolangs
 
 # recent containers should be updated when adding or modifying a language, so that
 # the travis build process will test it. The process cant test all languages
 # without timing out so this is required to get passed that issue.
-RECENT_CONTAINERS=node
+RECENT_CONTAINERS=esolangs
 
 ALL_CONTAINERS=${CONTAINERS} base
 
--- a/docker/esolangs.docker
+++ b/docker/esolangs.docker
@@ -1,0 +1,28 @@
+# Pull base image.
+FROM codewars/base-runner
+
+# Install "bf" Fast Brainf**k Interpreter
+RUN apt-get update && apt-get install -y --no-install-recommends bf
+
+# add the package json first to a tmp directory and build, copy over so that we don't rebuild every time
+ADD package.json /tmp/package.json
+RUN cd /tmp && npm install --production
+RUN mkdir -p /runner && cp -a /tmp/node_modules /runner
+
+# ADD cli-runner and install node deps
+ADD . /runner
+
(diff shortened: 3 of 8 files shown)


jhoffner:
@DonaldKellett this is up on preview.codewars.com. Will you please test? Please do not publish any kata as we don't want them showing up in production yet.

DonaldKellett:
@jhoffner Thanks for adding BF on the preview site 🎉   Sure, I'll test it right now :smile:

> Please do not publish any kata as we don't want them showing up in production yet.

No problem - I'll publish Kumite instead ATM

DonaldKellett:
@jhoffner I just tested BF support by publishing a few simple Kumite (nothing major, just Hello World and CAT programs).  It seems that BF execution is working just fine with or without program input but the "Preloaded" section is not being considered and executed (this issue is shown in my last Kumite which contains failed tests).  Perhaps I (or someone else) could add Preloaded code support for BF in my/his/her spare time through a separate PR?

DonaldKellett:
Bump.

@jhoffner @kazk When will BF finally be supported on the main site as an official Beta language (for Kata, not just for Kumite)?  There are at least two BF translations that I'd like to author but it's currently impossible to translate any Kata into BF.

PR #437, opened 2017-06-12
Lock ESLint version to 3.19.0

kazk:
ESLint 4.0.0 was released on 2017-06-11 and introduced breaking changes.
Use ESLint 3.19.0 for now.

- http://eslint.org/blog/2017/06/eslint-v4.0.0-released

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ node_js:
   - "6"
 
 install:
-  - npm install -g eslint
+  - npm install -g eslint@3.19.0
 
 services:
   - docker


kazk:
Travis is failing because the version of Dart changed and broke things :(
Dart was `1.23.0-1` yesterday (last successful build) and now it's `1.24.0-1`.

```
Failed to precompile den:den:
'package:convert/src/hex/encoder.dart': malformed type: line 20 pos 13: cannot resolve class 'ChunkedConverter' from 'HexEncoder'
    extends ChunkedConverter<List<int>, String, List<int>, String> {
```

image

See https://github.com/dart-lang/sdk/issues/29543

PR #441, opened 2017-06-25
Chapel

SLx64:
Hi,

I've added everything to run the Chapel programming language. (http://chapel.cray.com/) 

Error-Handling-Methods are currently not supported in the CW-2 test framework. I think there are some issues with first class functions.

The code is based on Python, Javascript, PHP and C++ files.

### DIFF
--- a/lib/runners/chapel.js
+++ b/lib/runners/chapel.js
@@ -37,16 +37,13 @@ module.exports.run = function run(opts, cb) {
       });
     },
 
-    // sanitizeStdErr: function(error) {
-    //   error = error || '';
-    //   return error.replace(/clang.*-std=c[^\s]+/g, '')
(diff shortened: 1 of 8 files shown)


lib/runners/chapel.js
```diff
@@ -0,0 +1,58 @@
...
+        runCode({name: executable, args: []});
+      });
+    },
+
+    testIntegration: function(runCode, fail) {
+      if (opts.setup) handleSetup(opts, dir);
+      const executable = path.join(dir, 'solution');
+      const solutionFile = util.codeWriteSync('chpl', [
+        cw2,
+        opts.solution,
```
jhoffner: missing support for `opts.setup`

SLx64: opts.setup is supported - see line 26 & 57 :)

PR #443, opened 2017-07-03
ordered list of installed packages

AnikaErceg:


jhoffner:
Thanks for the contribution. Looks like there is something wrong with the Crystal image (likely a resource change that the docker file depended on), so I'll need to get that fixed before I can merge this or anything else.

AnikaErceg:
Happy to contribute. Sometimes I wish I could do more than just order a list, but all in due time. :)
Good luck with the fix :D

PR #444, opened 2017-07-05
Added Node8 support + Mocha/Node versions

OverZealous:
- Added Node8 and updated Node 6 (LTS) to v6.11.0 (with legacy support for specifying 6.6.0)
- Switched to manually starting the Mocha global binary so we can swap Node versions with Mocha now, too
- The JS runner now throws a useful error if provided an invalid version
- Added a series of specs to ensure we are running the version of Node we expect


OverZealous:
@jhoffner or @kazk Looks like something might be wrong with Crystal, but it's not related to my changes. Given that it's not specifying the version when installing, it's possible this is due to a version upgrade. The issue seems to center around not outputting anything when it used to output something.

Also, we might need to up the timeouts on Chapel, it failed due to timeouts on me once, then passed the second time.

kazk:
> something might be wrong with Crystal

Apparently the package `crystal_0.23.0-1` (`crystal-0.23.0-1-linux-x86_64.tar.gz`) was built using Debian 8 and it's not compatible with LLVM installed on Ubuntu 14.04.
A possible fix improving compatibility was committed few days ago.

But we need to specify the version for anything we install while building the image.
Many of the images are installing the latest stable and can break any time :(
Something like the following should do for Crystal:

```dockerfile
RUN set -ex \
  && cd /tmp \
  && wget -q -O crystal.tar.gz https://github.com/crystal-lang/crystal/releases/download/0.21.1/crystal-0.21.1-1-linux-x86_64.tar.gz \
  && mkdir -p /opt/crystal \
  && tar zxf crystal.tar.gz -C /opt/crystal --strip-components=1 \
  && rm crystal.tar.gz \
  && ln -s /opt/crystal/bin/crystal /usr/local/bin/crystal
```

The version `0.21.1` is from README, but the version on Codewars might have been different depending on when the used image was last built.

> we might need to up the timeouts on Chapel

As far as I know, the Chapel compiler is not really fast unless you have a decent machine.
Travis build environment with `sudo` has 7.5 GB memory ~2, bursted cores according to their [documentation](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments). I don't know the options used for containers, but I think it'll take a while on `t2.micro` if it timeouts on Travis.

documentation/environments/node.md
```diff
@@ -19,7 +19,7 @@ The following plugins are included:
 - transform-class-properties
 - transform-flow-strip-types
 
-### 6.x Babel Plugins
```
kazk: Can you also update the paragraph under the "JavaScript" heading?

> Node 0.10.33 and 6.0 are currently supported, each with the option of transpiling using Babel JS. Node 6.0 without babel is the recommended option due to it supporting most ES2015 features natively, and because it does not incur the additional transpilation performance penalty that Babel JS causes.

OverZealous: Good catch! I searched for `6.x` and `6.6.0`, but missed that one.

OverZealous:
I agree we need to lock down the versions installed.  I wonder if we can use `apt-get` to specify the version? It looks like running `apt-get crystal=0.21.1` [installs a specific version](https://linux.die.net/man/8/apt-get), but I've never used that myself.

I can try to add the Crystal changes into this PR, if we want. Though, honestly, it should probably be a separate one since technically this PR didn't break it (and it shouldn't break if we don't rebuild the current version).

kazk:
> I wonder if we can use apt-get to specify the version? It looks like running `apt-get crystal=0.21.1` installs a specific version

I think `apt` needs to know *where* to get the specified version in order to do that.

Currently the image is built using the official `setup.sh` script:
```sh
# https://dist.crystal-lang.org/apt/setup.sh
apt-key adv --keyserver keys.gnupg.net --recv-keys 09617FD37CC06B54
echo "deb https://dist.crystal-lang.org/apt crystal main" > /etc/apt/sources.list.d/crystal.list
apt-get update
```

This only contains the latest stable release so `apt-get install crystal=0.21.1` will fail with error `E: Version '0.22.1' for 'crystal' was not found`.

I haven't found a way to install specific version of Crystal through `apt` and getting the versioned release [tarball](https://crystal-lang.org/docs/installation/from_a_targz.html) is easy.

I agree that this should be fixed through separate PR.

jhoffner:
@OverZealous these changes look great! Thanks. I'm holding all PR's ransom however until we get the Crystal issue fixed ;) I can try to take a look today if no one else has time, I'm going to be submitting a Java Gradle/Spring PR in a bit.

OverZealous:
I'm taking a look at Crystal right now.

PR #445, opened 2017-07-07
Support for julia 0.6

codeneomatrix:
- updated julia.docker (v.0.6.0)
- updated  julia_spec.js
- updated julia.yml
- updated Test.jl 
#### issue

```julia
it('should handle basic code evaluation', function(done) {
      runner.run({
        language: 'julia',
        code: 'print("42")'
      }, function(buffer) {
        expect(buffer.stdout).to.equal('42');
        done();
      });
    });
```
generates the following error when run the file julia_spec.js:

```julia
  1 failing

  1) julia runner .run should handle basic code evaluation:

      Uncaught AssertionError: expected '' to equal '42'
      + expected - actual

      +42

      at test/runners/julia_spec.js:12:34
      at lib/shovel.js:28:5
      at complete (lib/shovel.js:281:5)
      at ChildProcess.<anonymous> (lib/shovel.js:252:5)
      at maybeClose (internal/child_process.js:877:16)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:226:5)
```

### DIFF
--- a/lib/runners/julia.js
+++ b/lib/runners/julia.js
@@ -30,7 +30,10 @@ module.exports.run = function run(opts, cb) {
 
       runCode({
         name: 'julia',
-        args: ['-ie', runCmd]
+        args: ['-q', solutionFile],
+        options: {
+          cwd: juliaCodeDir,
+        }
       });
     }
   });
--- a/lib/runners/julia.js
+++ b/lib/runners/julia.js
@@ -12,7 +12,7 @@ module.exports.run = function run(opts, cb) {
       if (opts.setup) codeWriteSync('julia', opts.setup, juliaCodeDir);
       runCode({
         name: 'julia',
-        args: ['-P', ['push!(LOAD_PATH, "', juliaCodeDir, '", "frameworks/julia")'].join(""),
+        args: ['', ['push!(LOAD_PATH, "', juliaCodeDir, '", "frameworks/julia")'].join(""),
           '-e', opts.solution]
       });
     },
--- a/lib/runners/julia.js
+++ b/lib/runners/julia.js
@@ -9,11 +9,16 @@ module.exports.run = function run(opts, cb) {
   var juliaCodeDir = temp.mkdirSync('julia');
   shovel.start(opts, cb, {
     solutionOnly: function(runCode) {
+      codeWriteSync('julia', opts.code, juliaCodeDir, 'solution1.jl', true);
       if (opts.setup) codeWriteSync('julia', opts.setup, juliaCodeDir);
+
+      let runCmd1 = `
+              include("${juliaCodeDir}/solution1.jl")
+            `;
+
       runCode({
         name: 'julia',
-        args: [' ', ['push!(LOAD_PATH, "', juliaCodeDir, '", "frameworks/julia")'].join(""),
-          '-e', opts.solution]
+        args: ['-ie', runCmd1]
       });
     },
     testIntegration: function(runCode) {
@@ -30,10 +35,7 @@ module.exports.run = function run(opts, cb) {
 
       runCode({
         name: 'julia',
-        args: ['-q', solutionFile],
-        options: {
-          cwd: juliaCodeDir,
-        }
+        args: ['-ie', runCmd]
       });
     }
   });
--- a/lib/runners/julia.js
+++ b/lib/runners/julia.js
@@ -12,7 +12,7 @@ module.exports.run = function run(opts, cb) {
       if (opts.setup) codeWriteSync('julia', opts.setup, juliaCodeDir);
       runCode({
         name: 'julia',
-        args: ['-P', ['push!(LOAD_PATH, "', juliaCodeDir, '", "frameworks/julia")'].join(""),
+        args: [' ', ['push!(LOAD_PATH, "', juliaCodeDir, '", "frameworks/julia")'].join(""),
           '-e', opts.solution]
       });
     },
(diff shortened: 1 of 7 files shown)


kazk:
@codeneomatrix 
To make Travis build `julia-runner`, uncomment `- TEST_IMG=julia` in [.travis.yml](https://github.com/Codewars/codewars-runner-cli/blob/76db7f9254bfeb8d07db26966233b02060f11a7c/.travis.yml#L26).

lib/runners/julia.js
```diff
@@ -12,7 +12,7 @@ module.exports.run = function run(opts, cb) {
       if (opts.setup) codeWriteSync('julia', opts.setup, juliaCodeDir);
       runCode({
         name: 'julia',
-        args: ['-P', ['push!(LOAD_PATH, "', juliaCodeDir, '", "frameworks/julia")'].join(""),
+        args: ['', ['push!(LOAD_PATH, "', juliaCodeDir, '", "frameworks/julia")'].join(""),
```
kazk: Maybe that first empty string is causing the issue.

Why not `julia -q file` with `cwd` set to `juliaCodeDir` instead?
Is there a reason to include test framework in `solutionOnly`?

```js
runCode({
  name: 'julia',
  args: ['-q', opts.solution],
  options: {
    cwd: juliaCodeDir,
  }
});
```

kazk: Sorry I meant `solutionFile` not `opts.solution`.

```js
runCode({
  name: 'julia',
  args: ['-q', solutionFile],
  options: {
    cwd: juliaCodeDir,
  }
});
```

kazk: I read some docs and I see that `-P` was deprecated.
By removing that flag, the first argument that `julia` receives is now an expression ("'push!'"). I think this is the cause of the empty output issue.

kazk:
Travis is not building `julia-runner` because the `Makefile` doesn't contain target `julia`.

Looks like https://github.com/Codewars/codewars-runner-cli/commit/fae448996d64126e27c9d45b62b617b7f54f85b7 removed `julia` from `CONTAINERS`.
Adding `julia` back to the list should fix this.

test/runners/julia_spec.js
```diff
@@ -23,7 +23,7 @@ describe('julia runner', function() {
         `,
         testFramework: 'Test'
       }, function(buffer) {
-        expect(buffer.stdout).to.equal('("<PASSED::>Test Passed","<:LF:>\\n")');
+        expect(buffer.stdout).to.equal('("<PASSED::>Test Passed", "<:LF:>\\n")');
```
kazk: Why is the output inside parentheses?

```
("<PASSED::>Test Passed", "<:LF:>\n")
```

test/runners/julia_spec.js
```diff
@@ -4,15 +4,15 @@ var runner = require('../runner');
...
+    /*it('should handle basic code evaluation', function(done) {
       runner.run({
         language: 'julia',
         code: 'print("42")'
       }, function(buffer) {
         expect(buffer.stdout).to.equal('42');
         done();
       });
-    });
+    });*/
```
kazk: Re-enable this test and see if the suggested fix resolves the issue.

lib/runners/julia.js
```diff
@@ -12,7 +12,7 @@ module.exports.run = function run(opts, cb) {
       if (opts.setup) codeWriteSync('julia', opts.setup, juliaCodeDir);
       runCode({
         name: 'julia',
-        args: ['', ['push!(LOAD_PATH, "', juliaCodeDir, '", "frameworks/julia")'].join(""),
+        args: ['-P', ['push!(LOAD_PATH, "', juliaCodeDir, '", "frameworks/julia")'].join(""),
```
kazk: `-P` is not deprecated? I thought you removed it because it was deprecated. Can you clarify?

The suggested fix I meant was this:
```js
runCode({
  name: 'julia',
  args: ['-q', solutionFile],
  options: {
    cwd: juliaCodeDir,
  }
});
```

Write `opts.solution` to a file and set `solutionFile` to the path of the file.
Use `-q` so that Julia banner is not printed.
Code in `opts.setup` should be accessible from the solution using `include` or `require`.

Please correct me if I'm misunderstanding.

codeneomatrix: it works now

kazk:
@jhoffner 

I noticed that test outputs are in parentheses and asked a [question](https://github.com/Codewars/codewars-runner-cli/pull/445#discussion_r126246943), but got no response.

```
("<PASSED::>Test Passed", "<:LF:>\n")
```

The way code is executed is unconventional.
Similar to `node -e "require('file.js')"` instead of `node file.js`.

test/runners/julia_spec.js
codeneomatrix: That's how the program was
Just update the syntax to version 0.6

kazk: Thanks. I asked because you authored that a year ago? (https://github.com/Codewars/codewars-runner-cli/commit/c9fc0f22c615515c62ac4f0d9f7074549c63f845)
Apparently, you also wrote the test framework.
I'm pointing this out because codewars.com cannot use it unless the output format is correct.

codeneomatrix: I understand, what exactly should the exit be?

kazk: `<PASSED::>Test Passed`
You can see the [documentation](https://github.com/Codewars/codewars-runner-cli/blob/master/documentation/output_format.md) for other cases.

codeneomatrix: thank you very much

PR #446, opened 2017-07-07
Fixes Crystal by installing a specific version

OverZealous:
- Thanks to kazk for getting the majority of the work done on this.
- I had to add `shards` to the path binaries to get it to build.
- The `crystal` inside `embedded/bin` is not the same, and could not run the code. Very weird.

I also went ahead and doubled Chapel's timeout within the Docker build, simply to avoid further headaches. This doesn't affect Chapel itself anywhere else, only during the build test. Hopefully 10s is enough.

### DIFF
--- a/docker/chapel.docker
+++ b/docker/chapel.docker
@@ -33,5 +33,5 @@ ADD . /runner
 WORKDIR /runner
 USER codewarrior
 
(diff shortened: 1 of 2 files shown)


OverZealous:
Loooook at all the green checkmarks. Yeeeeessssss!

PR #447, opened 2017-07-10
Added Gradle/Spring support to Java

jhoffner:
- GroovyServ is no longer used. Gradle’s daemon is used instead. Unfortunately performance is about 800ms slower on avg.
- Ability to customize which packages get loaded
- Spring support
- Ability to code blocks into multiple files (all languages)
- Multi-file Java support
- Improved Java exception output
- Java build info is now presented within output


docker/java.docker
```diff
@@ -35,38 +35,62 @@ ENV PATH /usr/local/groovy/bin:${PATH}
 RUN mv groovyserv-1.1.0 groovyserv
 ENV PATH /usr/local/groovyserv/bin:${PATH}
```
OverZealous: Do you want to remove these lines?

OverZealous:
Looks really good!

kazk:
I'll start reviewing soon.

ESLint issues:
Maybe we should turn off `no-regex-spaces`.
image

image

lib/utils/split-files.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * splits a single string into multiple files. The initial content is the root content, the additional splits will
+ * show up within thier own "files" object within the result object returned.
+ * @param content The string to be split
+ * @param regex defaults to a regex that supports `// @config: split-file File.ext`
+ * @returns {{root: string, files: {}, splits: number}}
+ */
+module.exports = function splitFiles(content, regex = /^[ \t#|\/]* ?@config[: ] ?split-file (.*$)/gm) {
```
kazk: Why do we need this when there's `opts.files`?

frameworks/java/prewarm.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+echo "loading" > /workspace/prewarm.status
+
+# prewarm by starting the gradle daemon. Running an initial test build will also speed things up a bit
+cd /runner/frameworks/java && gradle --daemon --offline test
```
kazk: @jhoffner 
Have you looked into the amount of memory used by Gradle daemon?
I've read that it's relatively high and `t2.micro` has only 1 GB right?

jhoffner: Yeah honestly I don't even fully understand why we haven't run out of memory as it is, considering all of the containers we have and Groovyserv should take up a lot of memory as well, and its running pretty smoothly so far. I think Docker is able to manage RAM pretty well but I haven't dug into the inner workings much. It may be the reason why we have the occasional timeouts though. 

If I have to I may have to bump the ram up to t2.small. I would like to get this up in preview though and do some real world testing.

lib/utils/split-files.js
jhoffner: Codewars and Qualified do not yet support opts.files, only CodeStudio. Its just a way for us to convert the solution/setup/fixture format to a multi-file format when absolutely needed. Its mostly needed because Java is picky about file names.

frameworks/java/prewarm.sh
kazk: How bad was the performance of simple `javac` and `java`?
I can see that `jvm-runner` was slow because of Clojure+Leiningen startup, but I've been able to compile and run Java in decent speed for my side project.

I can just `docker run` and this finishes in few seconds (multiple files and Codewars output).
```sh
# Takes JSON from STDIN. `yaml r -j` converts YAML to JSON.
docker run --rm -i --memory=1GB java-runner:openjdk8 <<< $(yaml r -j test_junit.yml)
```

jhoffner: It was pretty bad. I think things went from an avg of 8 seconds down to 3 with Leiningen. We used to get a lot of users complaining about Java run times, which is what prompted us to try to find a better way. 

Now with Gradle it's more like 4 seconds, but also there is an initial 6 - 8 second warmup time - which shouldn't affect most users unless they happen to hit a container twice in a row before it is fully warmed up. 

Gradle obviously isn't being introduced for performance reasons but instead for extended java capabilities. We have a bunch of enterprise clients who need Spring support, and we eventually want to expand to supporting Android, which needs Gradle. Of course CW should be able to benefit from this as well - though framework challenges haven't become popular on there yet.

If this ends up causing performance issues then I may roll back some of these changes, but I wanted to see how well it runs on our staging cluster first. If things don't go well then we may have to prioritize persistent containers even higher than it is now.

> `docker run --rm -i --memory=1GB`

We never specify dedicated memory limits when starting containers, so they may all be competing with each other. Java is the only image that actually pre-warms anything, the others just sit there - so it might now suck up all of the memory that normally was left open before.

jhoffner: Using `top` on a local container, it looks like an extra 256mb of RAM is taken up once I start the Gradle daemon.

kazk: You can also start monitoring the memory usage of the container by
```
docker stats
# or
# docker stats --format "table {{.MemUsage}}"
```
and creating/using the container in another terminal session.

This will stream something like this (not running, started, doing some work, stopped):
```
CONTAINER           CPU %               MEM USAGE / LIMIT   MEM %               NET I/O             BLOCK I/O           PIDS
CONTAINER           CPU %               MEM USAGE / LIMIT   MEM %               NET I/O             BLOCK I/O           PIDS
e9973a8299eb        0.00%               0B / 0B             0.00%               0B / 0B             0B / 0B             0
CONTAINER           CPU %               MEM USAGE / LIMIT   MEM %               NET I/O             BLOCK I/O           PIDS
e9973a8299eb        0.12%               40.34MiB / 1GiB     3.94%               168B / 0B           0B / 0B             18
CONTAINER           CPU %               MEM USAGE / LIMIT   MEM %               NET I/O             BLOCK I/O           PIDS
```

docker/java.docker
```diff
@@ -31,42 +31,62 @@ RUN mv groovy-2.4.10 groovy
 ENV GROOVY_HOME /usr/local/groovy/
 ENV PATH /usr/local/groovy/bin:${PATH}
 
-# Setup GroovyServ
```
kazk: GroovyServ is still downloaded in lines 25-27.
Is Groovy still needed (lines 29-32)?

lib/runners/java.js
```diff
@@ -1,86 +1,197 @@
...
-        runCode({
-          name: 'groovyclient',
-          args: [
-            '-cp', opts.dir,
-            '-e', `import ${names.full}; ${names.full}.main(null);`,
-            '-Cauthtoken', 'groovy'
-          ]
-        });
-      });
+      util.codeWriteSync(null, opts.solution, `${opts.dir}/src/main/java`, names.file);
```
kazk: This is just
```js
util.writeFileSync(`${opts.dir}/src/main/java`, names.file, opts.solution, true);
```
because language parameter for `util.codeWriteSync` is `null`. `!config.moduleRegExs[null]` so `defaultFileName` is returned.

Also if we're using `fs-extra`, there is [`fs.outputFileSync`](https://github.com/jprichardson/node-fs-extra/blob/master/docs/outputFile-sync.md) that creates the directory if it doesn't exist already. So this becomes:
```js
fs.outputFileSync(`${opts.dir}/src/main/java/${names.file}`, opts.solution).
```

---

By the way, I never understood why `util.codeWriteSync` or shared `config.moduleRegExs` exists. Is there a reason to have those definitions in a shared module?

jhoffner:
I'm not sure how I feel about this `func-call-spacing` eslint rule.

kazk:
Maybe you're more familiar with the style used to write CodeStudio? Do you want to have the same configuration?
I personally don't mind any style because many rules can be fixed automatically. I also got used to no semicolons while learning VueJS.

---

`func-call-spacing` was about calling `func ()` right? Not `function func ()`. Isn't this also an error in VueJS or standardjs style?

lib/runners/java.js
jhoffner: Yeah I want to phase `writeFileSync` out. I can update this to not use it. 

I didn't initially add the codeWriteSync stuff, that came from someone else who used to contribute to the project, but the basic idea I guess was just to encapsulate the idea of having to name files dynamically based off of the contents of the code. I think this design also came around a time when there was more stuff in the config.js file (like shared code between CoffeeScript and JS), and people just felt compelled to use it more. I'm not sure if this really answers your question though.

The entire config.js needs a revamp. I don't know that it needs to really exist at all anymore actually. The utilities need a revamp too. I want to move each utility method into its own file and cleanup all of the usages, because right now writeFileSync and codeWriteSync are ambiguous and hard to reason about. 

I also want to clean shovel.js up because it sucks. In this PR I renamed `exec` to `spawnEx` to at least not make it completely silly (since `exec` is something different in Node). I would however like to rip most of this stuff out into their own utility files and clean things up a bit so that its easier to follow the logic. Also upgrading to Node 8 and using awaits would make things easier to follow.

jhoffner:
Yeah bringing things more into alignment with the client code (vuejs) would be helpful. I'm mostly pretty flexible with rules and don't really care much way one or another, but having to switch between conventions can be disorienting. 

I also got *mostly* used to no semi-colons from VueJS, though everytime I define a variable placeholder without a value I do this `let foo;` still.

Using no semi-colons with Node doesn't make sense though since there isn't a transformation process. I'm fine with that though, either way I tend to add them or forget them as it is. 

I think the func-call-spacing was giving me errors for both `var a = function (){}` and `it ("should...)"`. The former is more Vue rules like I think, the latter is probably me just wishing I was using Ruby whenever I write `describe` or `it`. Either way we probably just need to find something that is consistent between both.

OverZealous:
You are really bad about semicolons. I tried to enforce them on every line except the last line in a function (for inline-function readability) in Qualified, you didn't like that. 😛

lib/runners/java.js
kazk: Can you take a look at [kazk/codewars-runner](https://github.com/kazk/codewars-runner) when you have time?
That one is basically the core of this project trimmed down to minimum.
`examples/` contains definitions of pseudo-runners.
Hopefully we share the similar view and reuse some parts.

I had stopped working on them during your vacation because I had some question, but I forgot what that was :(

jhoffner:
> You are really bad about semicolons.

The Rubyist in me takes pride in that 😛

> every line except the last line in a function (for inline-function readability)

How do I do that? Do you mean this?
```javascript
function(){
  var foo;
}; //<---- Jake thinks this semi is wack
```

kazk:
I think we should use something predefined and just accept it. Don't customize or complain.

OverZealous:
I meant the rule allowed for

```js
someArray.forEach(function(v){
  console.log(v) // last statement doesn't require a semicolon
});
```
But everywhere else does (so all non-function, non-class statements).

OverZealous:
Eh, I don't know about accepting some externally defined set of ESLint rules blindly. There are some really lame rules hiding out there.

For example, I'm a strong believer in trailing commas in multiline arrays & object literals, enforcing `camelCase` and `===`, and even rarely allowing for curly-brace-free statements (`curly: 0`).

jhoffner:
> I think we should use something predefined and just accept it. Don't customize or complain.

As long as CR client, server and CLI are mostly aligned I'm fine. I'm pretty good with the `extends: 'standard'` rules that Vue uses by default, I disabled a few of the annoying OCD ones.

```
module.exports = {
  root: true,
  parser: 'babel-eslint',
  parserOptions: {
    sourceType: 'module'
  },
  extends: 'standard',
  // required to lint *.vue files
  plugins: [
    'html'
  ],
  // add your custom rules here
  'rules': {
    // allow paren-less arrow functions
    'arrow-parens': 0,
    // allow async-await
    'generator-star-spacing': 0,
    // allow debugger during development
    'no-debugger': process.env.NODE_ENV === 'production' ? 2 : 0,
    'spaced-comment': 0,
    'no-multiple-empty-lines': 0,
    'brace-style': 0,
    'valid-typeof': 0,
    'one-var': 0,
    'indent': 0,
    'quotes': 0,
    'comma-dangle': 0,
    'no-throw-literal': 0,
    'no-multi-spaces': 0,
    'padded-blocks': 0,
    'key-spacing': 0,
  }
}
```

Qualified is going to have its own rules and that's fine. No semi's for the client is fine too, since going back and adding them would be a total PITA.

lib/runners/java.js
jhoffner: @kazk this looks really good. You basically did everything that I just mentioned. I can create a Node 8 Upgrade branch so that we can start working on integrating these ideas in.

kazk:
I'm happy as long as the codebase is consistent. I'm also happy to adjust depending on other projects' styles because I understand switching between styles is painful.

I only suggested to use predefined one because I don't have strong beliefs on styles and disagreements here seems like personal taste?

> Gofmt's style is no one's favorite, yet gofmt is everyone's favorite.

OverZealous:
No worries, just a bit of friendly ribbing about regular "debates".

kazk:
> Using no semi-colons with Node doesn't make sense though since there isn't a transformation process.

JavaScript works fine without semicolons because of automatic semicolon insertion standardized in ECMAScript. Node (V8) is not an exception. For example, `npm` is written without it.
Just be careful because it can be surprising in some cases.
```ruby
foo = 0
[1,2,3].each {|x| puts x} # works
```
```js
foo = 0
[1,2,3].forEach(x => console.log(x))
// fails because this is interpreted as 0[1,2,3].forEach
```

jhoffner:
"doesn't make sense" was misleading. It's true that ASI is part of the spec, but it has issues, like the ones you pointed out. However Babel will add semicolons back into the code for you in a way that fixes the ASI issues - which is what I meant. 

https://medium.com/@kentcdodds/semicolons-in-javascript-a-preference-dd8fc8b80895

kazk:
As far as I know, Babel inserts semicolons to where ASI will since it works by parsing into JavaScript AST and using plugins to transform the AST.
image

ASI issues are caught by linters.
So if you want to use the same configuration from CodeStudio, we can do that.

Let me know if I'm still misunderstanding.
I just thought it'll make things easier for you. I noticed myself forgetting semicolons when I come back from Go and thought the same thing is happening to you switching between projects.

jhoffner:
Ah, Babel was supposed to fix these issues but I guess maybe not. 

> I just thought it'll make things easier for you. I noticed myself forgetting semicolons when I come back from Go and thought the same thing is happening to you switching between projects.

Yeah Ruby messes me up, though I still instinctively add semis in some cases. I'm fine either way, though if we keep them it would be nice to limit how picky it is. My only hesitation with going with removing them is just the effort involved, though I guess ESLint can auto fix it for us?

OverZealous:
I know my opinion doesn't count for much here, but I don't think we should remove them as a standard. ASI was not intended to enable not using semicolons as a general language design—it's a "helpful" feature for fixing typos more than anything. (If it was an intentional design, it would have eliminated the risks in the language spec.)

The fact that it can hide subtle typos and we have to rely on a ESLint catch them is enough to make it a bad habit. In some cases (inline functions—especially fat arrow functions), it's much easier to leave them off. On the other hand, I think they can help with readability. For example, making it clearer when something is a function expression and not a top-level function:

```js
function foo1() {
    // gets hoisted
}

let foo2 = () => {
    // does not get hoisted
};
```

Obviously in a short function, it's easy to see the declaration, but the extra semicolon makes the second stand out. It's just an example, though. You also have declared but undefined variable example above, and I think it's also easier to see the end of a long function chain (esp. in Promise chains):

```js
return somePromise()
    .then(() => {
        return foo;
    })
    .then(…)
    .catch(…);
```

But that's just my 2¢.

kazk:
Removing/adding should be as easy as changing the `semi` rule and doing `eslint --fix`.

I thought it might be worth adjusting the rules to reduce errors, but I don't know anymore. We should avoid spending too much time on this. Maybe just keep the rules and configure the editor to warn/fix on save or define pre-commit hook.

To be honest, I can't stand this repository's commit history and seeing commits like "fix eslint, fix eslint again, fix eslint" is what bothers me more. Can we start caring about history? Doing squash merge or start asking the contributor to cleanup before merge? Having good commit messages and clean history definitely helps a lot later acting as a documentation.

jhoffner:
I've never been a fan of squashing because it can cause problems and then suddenly I'm spending time fixing mine or helping someone else fix their commit issues. Honestly the verbose history has never bothered me, and I actually prefer separate commits for PR's, because each new commit acts as a status update. I don't have any experience using the new(ish) Github squash merge feature so we can start giving that a try and see how it goes.

As for the rules we use, semis, no semis... I don't really care. This is another thing that just doesn't bother me. After 17 years of coding in JS I don't think I've ever really had an issue arise because of a semicolon not being there, at least not in a way that didn't affect the code immediately. It's more important to understand how JS statements work and how to avoid the issues that the ASI creates in the first place. In terms of them increasing readability - I honestly think my mind completely ignores them so they have no affect on me. Maybe they used to matter to me more but after years of preferring Ruby as my main language, my mind just filters them out as noise. I would be fine keeping a loose rule on them. Use them, don't use them. Whatever. 

The above points are all personal. The bigger, more objective issue here is the lack of consistency across code bases and having to remember the specific rules that ESlint will throw up on for each. I'm all for the idea of keeping codebase styling consistent, but sometimes ESlint is just way too strict for me. Maybe we can loosen up a few rules and also start finding ways to make things more consistent between at least the Coderunner projects so that less context rules switching is needed.

> Maybe just keep the rules and configure the editor to warn/fix on save or define pre-commit hook.

Yeah maybe that would work too. A big part of the issue here is me. I suck at remember to run ESLint before I submit my code. I think I might have grown to rely on Travis too much for running our entire Qualified test suite (which takes too long for me to do it on my own machine each time).

OverZealous:
Just a tip: you can enable ESLint processing right within the JetBrains IDEs, so it highlights those mistakes automatically as errors as you type. Works well enough for me on Q, I actually forgot to enable it on the CLI. The nice thing is it picks up the project's config automatically.

I agree I'd rather relax the rules a bit, especially for less critical things.

kazk:
> I've never been a fan of squashing because it can cause problems and then suddenly I'm spending time fixing mine or helping someone else fix their commit issues.

I completely agree with this point. I did try helping but I don't want to spend anymore time.

> Honestly the verbose history has never bothered me, and I actually prefer separate commits for PR's, because each new commit acts as a status update.

Just to be clear, I do like verbose history. I just don't like insignificant commits appearing in the project history. I like separate commits in PR too, as long as each are actual changes that's significant to the project.
I commit a lot locally and many of them contains some notes to myself. I reorganize them into logical chunks with `git rebase -i` and summarize my changes before pushing. I do this because I'm really forgetful and I don't like writing documentation separately.
Maybe I care too much or I just think like this because I hadn't considered about other contributors enough.

---

For the rules, I'm ok to relax them because I only care about consistency.

jhoffner:
> Maybe I care too much or I just think like this because I hadn't considered about other contributors enough.

I'm happy that you care and as you start to dig more into our other projects (the ones I own specifically) you will notice more inconsistencies that you will want to fix. I'm totally cool with that. 

I think digging through closed PR's is another great way to get caught up on recent changes but having it in the commit history is also really great. I'll definitely start using the Squash on merge Github feature and see how that goes. Its nice to have the pointless "ESLint updates" commit on an open PR, to see that someone did try to fix it as a status update, but then on merge we can squash that away.

jhoffner:
> you can enable ESLint processing right within the JetBrains IDEs

Good to know, though looks like this requires turning off power saving mode. I don't use intellisense because it slows my machine down and usually all of the little UI things popping up distract me. Funny how things work, when I was a Visual Studio/.NET developer I thought my life depended on those features. Granted VS.NET was a hell of a lot more smoother/performant with its coding assistance.

I'll figure something out to improve my workflow...

OverZealous:
Yeah, it's definitely not a bad thing to care about these things, these are all about code & project quality, really.

Issue #418, opened 2017-05-04
TypeScript: ES5 target support


vguzev:
Is it possible to add TypeScript configuration with ES5 target?
We can't use getters/setters with current version of TypeScript 1.8.10...

For example when I try to run this kata
https://www.codewars.com/kata/590b85b745eaa415e10000b1/train/typescript
I'm getting errors:
../home/codewarrior/solution.ts(28,7): error TS1056: Accessors are only available when targeting ECMAScript 5 and higher.

In command line this can be done via command:
tsc task.ts -t ES5

kazk:
This should be possible.
I think the target is supposed to be `es6` already, but it's not working.
There's unused `/runner/tsconfig.json` with

```json
{
  "compilerOptions": {
    "target": "es6",
    "module": "commonjs"
  }
}
```

---

@jhoffner
TypeScript support needs some fix and updates (#380).
As far as I know, using [`@types` modules](https://www.npmjs.com/~types) will allow us to remove all the TypeScript stuff at the root of the project. And `@types` will be included in compilation by default so `///` comments can go away.
See the blog [The Future of Declaration Files](https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/).

jhoffner:
Sounds good. Would love to get this up-to-date. We will need to make sure to make <code>///</code> backwards compatible if it isn't aren't already with 2.0.

kazk:
...I didn't think about the compatibility issues with TypeScript 1.8 and 2.0+ :(
Maybe it's not as easy as I thought then. I don't know TypeScript much so I won't be able to do this anytime soon unless someone can let me know the requirements.

jhoffner:
> Maybe it's not as easy as I thought then

If we can support both 1.8 and 2.0 at the same time then we should be fine. I guess we could just manually copy the 1.8 files into the node_modules folder on the image (unless there is a better way to support two versions via npm).

kazk:
[docker/node.docker#L23-L24](https://github.com/Codewars/codewars-runner-cli/blob/05b31129d1a27e656224f51c9ffc8fe24a504234/docker/node.docker#L23-L24) installs the latest stable release and Codewars currently has `v2.2.1`.

```dockerfile
# Install TypeScript
RUN npm -g install typescript
```

```javascript
// Kumite
const cp = require('child_process');
console.log(cp.execSync('which tsc', {encoding:'utf8'}));
[...]

Issue #380, opened 2017-04-19
TypeScript: Update Typings


javatlacati:
While creating a [translation in Typescript](https://www.codewars.com/kumite/58f7ef7c7a5c280cd1000023/edit) I've got

> ../home/codewarrior/solution.ts(11,75): error TS2339: Property 'from' does not exist on type 'ArrayConstructor'.

but you can run the solution without problems from Typescript playground, so I suppose that your es6 typings are outdated or you are forcing output to an old version od ecmascript.

kazk:
This is actually `target` issue and should be resolved with #476 (using `languageVersion` `2.4/ES6`).

javatlacati:
That-s a gerat new but still it doesn't work on preview I still got

> ../home/codewarrior/solution.ts(9,35): error TS2339: Property 'from' does not exist on type 'ArrayConstructor'.

can you reopen the issue???

kazk:
I don't think it's deployed yet.

kazk:
preview.codewars.com still shows `1.8.10`
![image](https://user-images.githubusercontent.com/639336/28997501-138503b8-79cb-11e7-904f-0d0b19568f9b.png)

when deployed, it should show versions like `2.4/ES3` and `2.4/ES6`.

PR #450, opened 2017-07-11
Enabled Typescript in Karma & Angular 2

OverZealous:
- TypeScript can now be run using `karma_tdd` and `karma_bdd` with Mocha
- Angular 4 can be tested (with quite a bit of imported code)
- Split out some of the Karma logic so it can be reused in the TS runner. Since it's still highly specialized, I didn't move it to a shared module yet.
- Upgraded Angular 1.5 and added Angular 1.6 to the mix for now
- Increased the default timeout for running TS tests (it takes at least 6-7 seconds to run Typescript+Karma)


jhoffner:
@OverZealous looks good. My only main concern is will TypeScript 2 be compatible with all of our 1.8 Codewars content. It would be nice to be able to support both versions but thats probably going to require working with versions outside of NPM.

OverZealous:
Uh, we've got 2.x installed. If you run the runner directly, and do `tsc --version`, it's 2.4

OverZealous:
This is because we don't include versions when installing our global modules.

OverZealous:
That being said, right now we can sort-of support a different one for Karma, because the Typescript used in Karma is not the same as the globally installed one.

jhoffner:
Yeah no, I totally believe you that 2.4 is installed. I'm saying that since 1.8 is no longer installed, I hope the content will be compatible. Maybe I'm not understanding. 

@kazk this is a good example of where your versioned docker/directory structure would come in handy.

kazk:
I described this concern in https://github.com/Codewars/codewars-runner-cli/issues/418.

OverZealous:
I'm pretty sure the public one right now is running 2.4, simply because we run `npm install -g typescript` in the Docker image, without specifying the version.

I could be wrong, but the local copy within our `node_modules` directory was 1.x (due to being included with `typings`), and it threw an error on all of the existing (unchanged) files within `runner/typings`.

jhoffner:
> I described this concern in #418.

> I'm pretty sure the public one right now is running 2.4, simply because we run npm install -g typescript in the Docker image

ah right I forgot. OK well in that case, at least we don't need to be worried about existing content since it hasn't been a major problem yet (or is it? I'm still not fully up-to-date on the issues backlog). 

We should probably target the locally installed typescript executable though so that we can lock versions to at least 2.x.

OverZealous:
Yeah, I can do that, I'll set it to `@^2.4` for this PR, that's a good idea.

Technically we need to upgrade `typings`, too, but it had major breaking changes (including renaming `ambient` to `global`, and I could barely get through this update without breaking something. 😝

kazk:
> this is a good example of where your versioned docker/directory structure would come in handy.

Yeah, images tagged with language version reduces a lot of pain.

Also is the typing compatible? (#380)

OverZealous:
I didn't touch the typings here. I would not be surprised if they were way out of date. We really should be installing them correctly, but TypeScript isn't really my area, and I really had a lot of trouble hacking this work together.

OverZealous:
OK, just to be 100% sure, I ran a test on our active runners, it's definitely been 2.4.1 for a while.

OverZealous:
Ugh, the `object-curly-spacing` rule is lame. I can't see `{foo: 1, bar:2}` being better than `{ foo: 1, bar: 2 }`

kazk:
What is `/typings` in the project root?

I don't know TypeScript much either, but I have something like this:
```dockerfile
RUN npm install -g typescript@^2.4.x @types/node@^8.0.x
```

TypeScript compiled to JavaScript and executed.
```sh
tsc --target ES2017 \
    --module commonjs \
    --typeRoots /usr/local/lib/node_modules/@types \
    $ts-file

node $js-file
```

This way `/// <reference path="/runner/typings/node/index.d.ts" />` is no longer necessary.
I don't know about the compatibility.

jhoffner:
> Why is /typings in the project root?

This was a while ago and I forgot the specifics, but it might have something to do with NPM issues - and being related to why I needed to install using package.json instead of just using `npm install` commands directly within the Dockerfile. 

I think I might have struggled to get the paths to be clean as well. I didn't want users to have to know some cryptic path in order to find the reference, so I added them to the root.

kazk:
I'm getting really confused.
[`typings`](https://www.npmjs.com/package/typings) is "The TypeScript Definition Manager."?
How is this used? I don't see `typings install` in the Dockerfile.

jhoffner:
I think the typings were generated outside of Docker, due to NPM issues with generating them. I dont know if NPM still has Docker problems but back then it was extremely flimsy inside of Docker. 

So I generated them outside of the docker build process, which was a copout, and the reason why they have become so outdated. I see that I still install typings globally within the image, so I may have left that behind accidentally. I think I might be misremembering something as well - maybe NPM wasn't part of the issue IDK - but I remember this being a total PITA for some kind of reason. Unfortunately I didn't document the issues properly.

Either way, it's a hack and needs to be handled. We can handle that as a separate PR though.

kazk:
Ok, that makes sense. `npm install -gq typings` confused me.
I haven't experienced npm issues so it might have resolved since you experienced it.
Caching might have been painful since you do `ADD . /runner`?
The size of images can be easily reduced by removing `/tmp/node_modules` after copying to `/runner/node_modules` or just move instead.
Depending on your Docker version, you can also do multi-stage build which caches images separately so you don't have to do the hack to begin with.

PR #458, opened 2017-07-14
Ruby/BF/Chapel Updates

jhoffner:
- Ruby now has bcrypt gem, for better Rails integration
- BF now supports setup code
- Chapel setup code was reworked to not use mutation


jhoffner:
Thats odd, not sure why its picking up older merged PR's as a part of this.

kazk:
image

jhoffner:
The order in which I pull down changes doesn't usually matter, develop was behind master when I submitted this and I only pulled after I tried to submit a PR and it was showing recent changes as part of the actual changed files. When squashing do I need to make sure my branch is caught up to head before I commit anything?

kazk:
Maybe I misunderstood. 
I thought this was because you're reusing the `develop` branch containing those commits.

image

jhoffner:
Yeah I pretty much always use the develop branch for CR, since I usually work on one thing at a time. In this case I made those original changes into develop, pushed to origin, submitted a PR and merged into master. Those changes should therefore already be in both develop and master. I'm thinking maybe what happened is that somehow squashing those changes via the PR without immediately pulling back down to my local caused this PR to see the original, pre-squashed commits, as something new.

kazk:
Is there a reason use `develop` instead of feature branches (`feature/foo`)?

I make branch from `master` for each feature and `master` only merges.
I also never merge `master` into feature branch. If the master contains updates that feature branch requires, I rebase the feature branch so it originates from the latest `master`.
This keeps things simple and easy to see how/when the feature was added.

image

I can also have `dev` branch in between that I develop against and have `master` merge `dev` for each version bump. This makes it easier to do DevOps stuff.

jhoffner:
We use feature branches for Qualified, though in CR the habit is not as well formed. It's never been an issue with using normal merge commits, but I guess the simplified workflow is not compatible with squashes. 

The reason why I don't silo things into feature branches as much with CR (and also CW) is that my development role tends to often be spent being the glue that sticks things together. I work with 3 or more projects a day and am often bouncing back and forth, handling multiple things. I don't usually sit down and go, OK, I'm about to work on feature X, lets start working on that. Its more, OK im trying to get this work and oh look, CR needs an update, ok now while that is taking 5 minutes to build I'm going to switch to CW, it has something that needs to be updated because of the CR update and oh yeah while I'm in here I'll fix that bug that I see right in front of me... 

Anyway my point here is that utilizing squashing seems to require a less fragmented and disciplined branching workflow for it to not cause issues. I'm all for improving my own workflow and being more disciplined, it certainly helps towards enforcing the community to do the same, however this was kind of my initial gripe with squash/rebase - if you are not extremely disciplined and on top of things, and everyone on your team the same, then you end up spending a lot of time going back and trying to clean things up.

kazk:
I read some documentation and the source of the problem seems like squash taking a group of commits and creating a new commit. So the original commits are not recognized as merged, showing up in this PR.

Thanks for explaining, I understand the difficulty better now.

PR #459, opened 2017-07-14
Fix/warnings stderr

jhoffner:


jhoffner:
Ugh, @kazk it did it again. I'm, going to find a reason to change something else so I can test something out. I was going to add this to my .zshrc

```
# Create a new Git Feature Branch. Ensures that you set to master (or develop if thats the flow),
# pulls latest, then creates a new branch of off master and checksout
# usage: fb origin master feature/example
function fb {
	git checkout $2
	git pull $1 $2
	git branch $3
	git checkout $3
}
```

I might also add auto-stashing to this and then a switch branch method which would automatically re-apply any stashed changes.

kazk:
I think you need to apply the changes after `fetch`.

```sh
# `git fetch` only updates the refs
# `git pull` does `merge` after `fetch`
# `git fetch` followed by `git merge` might be less magical
function gfb {
	git checkout $2
	git pull $1 $2
	git checkout -b $3 # `-b` can make you a branch
}
```

kazk:
@jhoffner
I'm starting to think using a squash merge to keep the project history clean wasn't a good idea.
I'd rather live with insignificant commits than disoriented trees. I didn't the consequences of changing the history :(
I think I'll have the same issue as you had in this PR if I rebase a feature branch that builds on top of another feature branch (worked on while waiting for the first feature branch to be merged).

Let's stop using it if you're okay. And only use it in extreme cases when we get a contribution with too much noise from a contributor who doesn't know git to fix?

jhoffner:
Works for me. Feel free to rebase/squash locally but Github's squash feature just seems to cause more problems than it solves.

Issue #240, opened 2016-09-18
Random tests are currently impossible in Rust


error256:
I'm not really sure but it looks like there's no way to generate random numbers properly without using an external crate which requires adding a dependency to Cargo.toml project file.
https://doc.rust-lang.org/rand/rand/index.html
https://stackoverflow.com/questions/30735490/unresolved-name-randthread-rng

lteacher:
Hi,

Originally there was a solution which used cargo but it was more irritating for some various reasons and now just `rustc` is used. It looks like either way an enhancement is required to do it.

DonaldKellett:
*Bump*

@jhoffner Do you know an active contributor to this Repo who might just be able to allow for random input generation in Rust?  I have just received a Rust translation to one of my Kata without random tests and I would love to approve it (the overall translation quality is good) but I can't until he/she has added random tests to it.

PR #464, opened 2017-07-15
Set Rust version to 1.15.1

kazk:


lteacher:
Hi,

I got a notification from the issue #240 and figured to check it out since I added it originally. Anyways I was looking into it and of course noticed your issue and links to the issue.

I just wanted to post here at your PR because I know that to fix the random tests issue it would probably mean changing to use Cargo to get the rand extern crate in. Since you are adjusting the docker image etc here its prob a good time for me to appear and mention it. 

Maybe instead of specifically installing the version like this we can use [`rustup`](https://www.rustup.rs/) as it will correctly include `cargo`. Disclaimer though, I know less about rust than the small amount I knew when I added support back whenever that was... but i guess instead you will use `rustup install 1.15.1` to get the fixed version. This would likely help to allow for multiple versions.

docker/rust.docker
```diff
@@ -1,28 +1,38 @@
 FROM codewars/base-runner
 
-# Install Rust Compiler
-RUN curl -sSf https://static.rust-lang.org/rustup.sh | sh
+# Install Rust v1.15.1 to /usr/local
+RUN set -ex \
+ && cd /tmp \
+ && curl -fSsL https://static.rust-lang.org/dist/2017-02-08/rust-1.15.1-x86_64-unknown-linux-gnu.tar.gz -o rust.tar.gz \
```
lteacher: Looks like this would get cargo after all? still though the rustup did seem a little more flexible, looks something like this average version I attempted:

```dockerfile
RUN cd /tmp \
    && curl https://sh.rustup.rs -sSf -o rustup-init.sh \
    && chmod +x rustup-init.sh \
    && sh rustup-init.sh -y --default-toolchain 1.15.1 \
    && rm rustup-init.sh
```

kazk: Thanks! This is definitely better. I didn't read the `rustup` documentation enough. I'll take a look at your PR.

kazk:
@jhoffner
I have `rust-runner` using `cargo` that has `rand` installed ready to be reviewed.
I think it's better to open a separate PR than adding them here. It's a major change and I want to make it clear. I'd also like some opinions for the hacks I did in order to maintain backward compatibility.
Can you merge this? I don't think this one will cause any issues since it's just specifying the version.

PR #465, opened 2017-07-15
Fix Swift compatibility issue around XCTMain

### LINKED ISSUE #463: New Swift Runner is not rendering proper output
@kazk I have no idea what is going on but the latest runner is not sending proper output:

https://preview.codewars.com/kata/some-egyptian-fractions/train/swift

image

I'm looking into it now to make sure there wasn't some sort of deployment error but I need to leave soon.

kazk:
This should resolve #463

### DIFF
--- a/lib/runners/swift.js
+++ b/lib/runners/swift.js
@@ -52,7 +52,7 @@ module.exports.run = function run(opts, cb) {
         args.push('/workspace/setup.swift');
       }
       fs.writeFileSync('/workspace/solution.swift', opts.solution);
-      fs.writeFileSync('/workspace/main.swift', opts.fixture);
+      fs.writeFileSync('/workspace/main.swift', opts.fixture.replace(/^\s*XCTMain\(/m, '_XCTMain('));
       args.push('/workspace/solution.swift', '/workspace/main.swift');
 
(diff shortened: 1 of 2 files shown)


lib/runners/swift.js
```diff
@@ -52,7 +52,7 @@ module.exports.run = function run(opts, cb) {
         args.push('/workspace/setup.swift');
       }
       fs.writeFileSync('/workspace/solution.swift', opts.solution);
-      fs.writeFileSync('/workspace/main.swift', opts.fixture);
+      fs.writeFileSync('/workspace/main.swift', opts.fixture.replace(/^\s*XCTMain\(/m, '_XCTMain('));
```
jhoffner: We may or may not want to support both with and without a leading `_`. We should probably support whitespace in between `n(` just in case.

kazk: Added whitespace flexibility, but no leading `_` allowed. There shouldn't be any content using `_XCTMain` yet right?

kazk:
node-runner is failing again. Probably MongoDB test case.
I hope this fixes the issue for now.
Just wanted to let you know that I might not be available until tomorrow.

PR #466, opened 2017-07-17
Rewrite Rust runner

### LINKED ISSUE #240: Random tests are currently impossible in Rust

### LINKED ISSUE #462: Specify Rust version and use cargo
- [x] Specify version (Set to 1.15.1 in #464)
- [x] Support random testing (#240)
- Look into the timeout issue (#440). See if this issue is reproducible after #466.

image

kazk:
This should allow writing random tests using `rand` (#240).
I had to come up with a backward compatible way, and I'd like feedbacks from Rust users.

### Features

- Use Cargo
  - Install `rand` package
- Improve output
  - Align with other languages
  - Include stdout of failed tests
  - Proper termination

image
image

### Other Changes

- Include examples to tests
- Restructure tests by context

### Structure

Needs some hack in order to maintain backward compatibility and avoid errors with duplicate import.

    .
    |- Cargo.toml
    `- src
       |- lib.rs    (opt.setup + opts.solution + module declaration)
       `- tests.rs  (opts.fixture)

- Placing `opts.fixture` in submodule `tests.rs` should avoid issues with duplicate imports
- Maintains backward compatibility by prepending `use super::*;` to `opts.fixture` if missing

Another idea was

    .
    |- Cargo.toml
    `- src
       |- lib.rs      (module declarations)
       |- solution.rs (opts.setup + opts.solution)
       `- tests.rs    (opts.fixture)

Decided against this because maintaining backward compatibility is hard.
- Requires solutions to be declared public
- Requires tests to import solution `use solution::*;`

### Using `rand` in tests

    extern crate rand;
    use self::rand::Rng;

Needs `self::` because `extern crate` loads to current name-space and tests are in `tests` module.

---

Resolves #240
Closes #462

### DIFF
--- a/test/runners/rust_spec.js
+++ b/test/runners/rust_spec.js
@@ -1,8 +1,18 @@
+"use strict";
+
 const expect = require('chai').expect;
 const runner = require('../runner');
+const exec = require('child_process').exec;
 
 describe('rust runner', function() {
-  describe('.run', function() {
+  describe('running', function() {
+    afterEach(function cleanup(done) {
+      exec('rm -f /workspace/rust/src/*.rs', function(err) {
+        if (err) return done(err);
+        done();
+      });
+    });
+
     it('should handle basic code evaluation', function(done) {
       runner.run({
         language: 'rust',
@@ -13,20 +23,6 @@ describe('rust runner', function() {
         `
       }, function(buffer) {
         expect(buffer.stdout).to.equal('Bam\n');
-        done();
-      });
-    });
-
-    it('should handle invalid code', function(done) {
-      runner.run({
-        language: 'rust',
-        code: `
-          fn main() {
-            println!("Bam);
-          }
-        `
-      }, function(buffer) {
-        expect(buffer.stderr).to.contain('unterminated double quote');
         done();
       });
     });
@@ -44,12 +40,25 @@ describe('rust runner', function() {
             thread::spawn(move|| {
                 tx.send(10).unwrap();
             });
-    
+
             println!("{}", rx.recv().unwrap());
           }
         `
       }, function(buffer) {
         expect(buffer.stdout).to.contain(`10`);
+        done();
+      });
+    });
+  });
+
+  describe('examples', function() {
+    runner.assertCodeExamples('rust');
+  });
+
+  describe('testing', function() {
+    afterEach(function cleanup(done) {
+      exec('rm -f /workspace/rust/src/*.rs', function(err) {
+        if (err) return done(err);
         done();
       });
     });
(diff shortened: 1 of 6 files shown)


kazk:
@lteacher

> Originally there was a solution which used cargo but it was more irritating for some various reasons 

What prevented you to use `cargo`? Did I miss anything?
Let me know if I can improve this somehow.

@error256, @darikg, and any other Rust users

I don't know Rust and this was the only way I could think of while maintaining backward compatibility with existing contents. Let me know what you think.

lteacher:
@kazk Yeah I think it was just back then I overlooked the `rand` stuffs and seemed that using cargo was a bit more annoying thats all.

test/runners/rust_spec.js
```diff
@@ -103,165 +113,352 @@ describe('rust runner', function() {
...
         language: 'rust',
-        setup: `
-          use std::thread;
-          use std::sync::mpsc::channel;
+        code: 'fn add(x: i32, y: i32) -> i32 { x + y }',
+        // need `self::` because `extern crate` loads to current namespace and tests are in `mod tests`.
+        fixture: `
+          extern crate rand;
+          use self::rand::Rng;
+          use self::rand::distributions::{IndependentSample, Range};
```
lteacher: At a glance, it seems unfortunately that we must use this. Would be good to find a way without it before it goes in since once its in will be kinda locked I guess? Unfortunately I can't suggest an alternative just at a glance as I'm not proficient at rust but am happy to tinker a little bit later (Melbourne time :wink:) to see if there is any help I can provide.

kazk: Thanks for reviewing.
Yeah, that's bothering me too.
I also tried having `opts.setup` like
```rust
#[cfg(test)] extern crate rand;
```
But this obviously breaks when `opts.solution` has `extern crate rand;`.

I used [Test Organization](https://doc.rust-lang.org/book/second-edition/ch11-03-test-organization.html) chapter from "The Rust Programming Language" as a guide.

---

Currently we can only have one version of "rust-runner" so we need to be careful not to break existing contents.

> am happy to tinker a little bit later (Melbourne time 😉) to see if there is any help I can provide.

Thanks :)

lteacher: Actually went back over it and think it looks really good. Unfortunate that cargo was not used originally.

I can see why this is as above, it seems likely people will write the tests though assuming that they won't need the `self` on there. At least its probably not that likely for solutions to contain the rand crate generally.

PR #469, opened 2017-07-18
Add R support

kazk:
R 3.4.1, test integration with [testthat](https://github.com/hadley/testthat).
No other packages are installed for now. A package installed as testthat's dependency should not be considered available. Package requests can be made by opening an issue.

- <https://www.r-project.org/>
- <https://github.com/hadley/testthat>


jhoffner:
@kazk this is awesome! I have been wanting R on here for a long time. The implementation looks good to me. Only thing I see missing is the updated readme.

kazk:
Glad to know that :) I think we can have interesting outputs using packages, but I haven't looked into them yet.

Can you merge #467? I skipped Travis using `[ci skip]` and apparently GitHub doesn't know that.
I'll rebase this with README update on top of those.

PR #482, opened 2017-07-31
[WIP] Reenable sql_spec.js

kazk:
SQL tests are passing now. OCaml runner is failing for some reason :/

### DIFF
--- a/docker/ruby.docker
+++ b/docker/ruby.docker
@@ -191,7 +191,7 @@ ENV USER=codewarrior HOME=/home/codewarrior
 
 RUN mocha -t 5000 test/runners/ruby_spec.js
 RUN mocha -t 5000 test/runners/shell_spec.js
-#RUN mocha -t 5000 test/runners/sql_spec.js
+RUN mocha -t 5000 test/runners/sql_spec.js
 RUN sh /runner/lib/cleanup.sh
 
 ENTRYPOINT ["./entrypoint.sh"]


kazk:
I'm going to open a PR fixing OCaml. I don't know the exact cause of this failure, but I was able to build by locking the versions and using external package solver.

kazk:
OCaml is passing now :/ Maybe the issue was related to GitHub outage?
#483 is still relevant because it locks versions.

jhoffner:
Are you sure they are working consistently? I think they work only part of the time.

kazk:
I think I tried few times. I'll look into this again.

kazk:
@jhoffner 

SQL tests passes for building, but fails when I try **after** the build.

```
docker run --rm -it --entrypoint mocha codewars/ruby-runner -t 5000 test/runners/sql_spec.js
```

```
  1) sql runner .run solution only should support postgres:
     Uncaught AssertionError: expected '<STATUS::>Connecting to database...\n' to include '"name":"c"'
      at test/runners/sql_spec.js:59:36
      at lib/shovel.js:28:5
      at complete (lib/shovel.js:297:5)
      at ChildProcess.<anonymous> (lib/shovel.js:268:5)
      at maybeClose (internal/child_process.js:877:16)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:226:5)
```

When I bash into `ruby-runner` and execute `mocha -t 5000 test/runners/sql_spec.js` for the first time, the test case `"should support postgres"` fails.

```
<STATUS::>Connecting to database...
/usr/local/bundle/gems/sequel-4.49.0/lib/sequel/adapters/postgres.rb:225:in `initialize': PG::ConnectionBad: FATAL:  the database system is starting up (Sequel::DatabaseConnectionError)
	from -e:5:in `<main>'
<STATUS::>Connecting to database...
```

So far, only the first time of the session fails.

kazk:
@jhoffner I just built `ruby-runner` from scratch and tests are passing, even when I run them after building. The problem is that because I built from scratch and versions of gems are not locked, many are updated including Sequel. Sequel is now 5.0.0 and the deprecation warning in #501 is now error because the feature was removed.
I don't know if the failures we had was related to Sequel's versions. I'm not sure what to do.
Tested many times like the following:
```shell
for _ in $(seq 1 10); do docker run --rm -it --entrypoint mocha codewars/ruby-runner -t 5000 test/runners/sql_spec.js; done
```

PR #487, opened 2017-08-04
[WIP][RFC] Add Kotlin support

### LINKED ISSUE #165: Add support for Kotlin
Kotlin is this new language created by the JetBeans company. It's a statically typed language for Android development working on the JVM. The language is influency by many languages such as Java, Groovy and C#. It has a lot of potential and I think a language, open source might I've add (https://github.com/JetBrains/kotlin), will be a fine addition for Codewars.

kazk:
Kotlin 1.1.3-2, tests with JUnit 4.12.

Tried to make a clean and simple starting point. Most of the work is defined in `build.gradle` and this simply does `gradle run/test`. Maybe java-runner can do something similar to simplify things?

- code can be in default unnamed package or arbitrarily named package
- tests can have multiple suites
- should be easy to extend this to support `opts.files`, just output files to correct locations
- `KOTLIN_GRADLE_DAEMON` environment variable decides when to use Gradle daemon

---

Closes #165

### DIFF
--- a/docker/kotlin.docker
+++ b/docker/kotlin.docker
@@ -1,0 +1,58 @@
+FROM codewars/base-runner
+
+# Install Oracle JDK 8
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends software-properties-common \
+ && add-apt-repository ppa:webupd8team/java \
+ && apt-get update \
+# http://askubuntu.com/a/190674
+ && echo debconf shared/accepted-oracle-license-v1-1 select true | debconf-set-selections \
+ && echo debconf shared/accepted-oracle-license-v1-1 seen true | debconf-set-selections \
+ && apt-get install -y --no-install-recommends oracle-java8-installer
+
+# Install Gradle
+RUN apt-get update && apt-get install -y --no-install-recommends zip unzip
+ENV GRADLE_HOME=/usr/local/gradle \
+    GRADLE_VERSION=4.0 \
+    GRADLE_DOWNLOAD_SHA256=56bd2dde29ba2a93903c557da1745cafd72cdd8b6b0b83c05a40ed7896b79dfe
+RUN set -o errexit -o nounset \
+	&& echo "Downloading Gradle" \
+	&& wget --no-verbose --output-document=gradle.zip "https://services.gradle.org/distributions/gradle-${GRADLE_VERSION}-bin.zip" \
+	\
+	&& echo "Checking download hash" \
+	&& echo "${GRADLE_DOWNLOAD_SHA256} *gradle.zip" | sha256sum --check - \
+	\
+	&& echo "Installing Gradle" \
+	&& unzip gradle.zip \
+	&& rm gradle.zip \
+	&& mv "gradle-${GRADLE_VERSION}" "${GRADLE_HOME}/" \
+	&& ln --symbolic "${GRADLE_HOME}/bin/gradle" /usr/bin/gradle
+
+RUN ln -s /home/codewarrior /workspace
+ENV NPM_CONFIG_LOGLEVEL warn
+
+WORKDIR /runner
+COPY package.json package.json
+RUN npm install --production
+
+COPY frameworks/kotlin frameworks/kotlin
+RUN chown -R codewarrior:codewarrior frameworks/kotlin
+
+USER codewarrior
+ENV USER=codewarrior HOME=/home/codewarrior
+
+RUN cd /runner/frameworks/kotlin \
+ && gradle --no-daemon test || true
+
+COPY *.js ./
+COPY lib/*.js lib/
+COPY lib/*.sh lib/
+COPY lib/utils lib/utils
+COPY lib/runners/kotlin.js lib/runners/
+COPY examples/kotlin.yml examples/
+COPY test/runner.js test/
+COPY test/runners/kotlin_spec.js test/runners/
+
+RUN mocha test/runners/kotlin_spec.js
+
+ENTRYPOINT ["node"]
--- a/frameworks/kotlin/build.gradle
+++ b/frameworks/kotlin/build.gradle
@@ -1,0 +1,75 @@
+buildscript {
+  ext.kotlin_version = '1.1.3-2'
+  repositories {
+    jcenter()
+  }
+  dependencies {
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+  }
+}
+
+apply plugin: 'kotlin'
+apply plugin: 'application'
+
+mainClassName = System.getenv('KOTLIN_MAIN_CLASS_NAME')
+defaultTasks 'run'
+
+repositories {
+  jcenter()
+}
+
+dependencies {
+  compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+  testCompile 'junit:junit:4.12'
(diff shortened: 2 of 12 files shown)


docker/kotlin.docker
```diff
@@ -0,0 +1,58 @@
+FROM codewars/base-runner
```
jhoffner: Lets reuse the `java.docker` image for now (eventually we will call this jvm, once we fully replace the existing one). 

At some point managing different versions may be an issue with Kotlin + Java, but for now having to manage a runner container instance for each JVM language we support is going to cost us more RAM and pre-warm/startup time on the server.

docker/kotlin.docker
```diff
@@ -0,0 +1,58 @@
...
+ && gradle --no-daemon test || true
+
+COPY *.js ./
+COPY lib/*.js lib/
+COPY lib/*.sh lib/
+COPY lib/utils lib/utils
+COPY lib/runners/kotlin.js lib/runners/
+COPY examples/kotlin.yml examples/
+COPY test/runner.js test/
+COPY test/runners/kotlin_spec.js test/runners/
```
jhoffner: In order for prewarming to work for Gradle, you need to include `ADD frameworks/java/prewarm.sh /runner/prewarm.sh`

frameworks/kotlin/build.gradle
```diff
@@ -0,0 +1,75 @@
+buildscript {
```
jhoffner: Hmmm. I might see why you decided to split this out. Prewarming one gradle build files on container start may prevent the code from building efficiently if you switch the build during code execution. 

Tricky to say. In other words, If we prewarm the image with the java gradle file, then run the code later with the kotlin, the prewarmed gradle instance may end up not being used and a new one would be used instead.

frameworks/kotlin/build.gradle
```diff
@@ -0,0 +1,75 @@
...
+  }
+}
+
+test {
+  reports {
+    html.enabled = false
+    junitXml.enabled = false
+  }
+
+  beforeSuite { desc ->
```
kazk: I decided to split for now because I didn't want to break Java and I also wanted to use these listeners in `build.gradle`. This should allow adding test integrations for other JVM languages easy. I think Groovy and Scala can be supported on top of this by just adding plugins.

docker/kotlin.docker
kazk: Is there a reason to use separate shell script? I added few lines in `listen.js`.

jhoffner: Just separation of concerns. listen.js otherwise starts to become a mixmash of different runner logic.

frameworks/kotlin/build.gradle
jhoffner: OK, we can keep it separate for now. I also wanted to update the java version to use listeners at some point. I think the only thing that is special to the CwRunListener is the `<LOG::-Exception Details>` bit, which probably also be doable within the listeners.

kazk:
I'm concerned about the memory usage as well. I usually like to have single purpose images, but I can see that having each image for JVM languages is painful. I'll look if we can have `gradle-runner` that can switch languages/test frameworks smoothly.

kazk:
I'll rebase this soon.

kazk:
On GCP `n1-standard-1` (1 vCPU, 3.75 GB Mem), I got these stats.

```
NAME                MEM USAGE / LIMIT
kotlin              149.3MiB / 3.611GiB
java                153.8MiB / 3.611GiB
ruby                8.234MiB / 3.611GiB
```

```shell
docker run -d --name kotlin codewars/kotlin-runner listen
docker run -d --name java codewars/java-runner listen
docker run -d --name ruby codewars/ruby-runner listen

docker stats --format "table {{.Name}}\t{{.MemUsage}}"
```

When `kotlin/java-runner` starts, the memory usage goes up to ~220 MB and eventually drops to ~150 MB and stays. Other images should be similar to ruby.

frameworks/kotlin/build.gradle
```diff
@@ -0,0 +1,75 @@
+buildscript {
+  ext.kotlin_version = '1.1.3'
```
ice1000: The newest is `1.1.3-2`

frameworks/kotlin/src/main/kotlin/helloWorld.kt
```diff
@@ -0,0 +1,14 @@
+package codewars
+
+// https://github.com/JetBrains/kotlin-examples/blob/master/gradle/hello-world/src/main/kotlin/demo/helloWorld.kt
+fun getGreeting(): String {
+    val words = mutableListOf<String>()
+    words.add("Hello,")
+    words.add("world!")
+
+    return words.joinToString(separator = " ")
+}
```
ice1000: A more kotlinic way:

```kotlin
val greeting: String get () = listOf("Hello,", "world!").joinToString(separator = " ")
```

The above code generates the equivalent jvm bytecode

kazk:
@ice1000 Thanks for reviewing. I'll update Kotlin version.
Do you know if we can have multiple test frameworks in one project? I'd like to use listeners defined in `build.gradle` if possible.

lib/config.js
```diff
@@ -15,15 +15,15 @@ module.exports = {
     go: 15000,
     haskell: 15000,
     sql: 14000,
-    java: 20000
+    java: 20000,
+    kotlin: 20000,
```
javatlacati: A very common complain about kotlin is that it compiles slower than Java. Pehaps you could consider adding some morer time.

javatlacati:
@kazk Perhaps you can add the test frameworks as test dependencies i.e TestNG, Junit, Mockito, Jasmine. Dependencies will not crash due to differrent package names.

lib/config.js
kazk: I know Kotlin takes longer to compile, but I'm not sure how much longer. It really depends on the complexity of the target. On Travis, minimum challenges finish in ~5 sec with Gradle daemon running. I don't know how long it'll take on AWS t2.small with other containers.
We can adjust later if this is not enough. We can also adjust by dependencies used like Java runner does. This is just the default.

kazk:
If adding support for other test framework is possible without changing the project structure, I'll leave this PR as is and let others contribute. I guess it can be done by adding test dependencies and adjusting `testIntegration` function.

kazk:
Actually I'm going to look into Gradle little bit more.

kazk:
I think I will be opening a new PR for `gradle-runner` with Kotlin/Groovy/Scala support instead.

- Kotlin
  - JUnit, optionally using assertion frameworks like Kluent, Expekt, Hamkrest
  - KotlinTest
  - Maybe Spek once Gradle has better support for JUnit 5
- Groovy
  - JUnit
  - Spock
- Scala
  - JUnit, optionally using ScalaTest's assertions/matchers

ice1000:
Sounds awesome.

Also, spek supports junit4 too. You can refer to [this gradle file](https://github.com/lice-lang/lice-repl/blob/master/build.gradle#L56) to find out how to use spek with junit4.

kazk:
@ice1000 Yeah, Spek itself is compatible with JUnit 4, but `gradle test` can't output properly as far as I know. I think it's necessary to write custom class implementing `TestExecutionListener` which is experimental, then use JUnit 5's launcher to register that. This also means that we need to write some custom Gradle task to handle it.

All of the ones I listed above can be supported by just adding the dependency :)

kazk:
Superseded by #490

PR #488, opened 2017-08-05
Add compile tools from jvm as dependency

javatlacati:
For very high level katas it can enable to access all of the set of compiling tools programatically.

Some ideas of usages are:

* create Katas to test you junit tests for a preloaded java class
* perhaps a Kata about creating your own progamming language that runs on the Java Virtual Machine
* create tests that actually detects trashy code ( e.g unused imports, unused methods )
* Run [Dynamic code](http://www.javaworld.com/article/2071777/design-patterns/add-dynamic-java-code-to-your-application.html)
* create katas to test Javadoc i.e. the quality of generated documentation
* Creation of custom annotations??? ( an annotation processor class should be addeded to gradle build task but.. you could personalize how the compiler works  )
* create your custom jvm runner kata

### DIFF
--- a/frameworks/java/build.gradle
+++ b/frameworks/java/build.gradle
@@ -32,6 +32,7 @@ dependencies {
   compile 'org.springframework.boot:spring-boot-starter-data-jpa:1.5.4.RELEASE'
   compile 'org.springframework.boot:spring-boot-starter-data-rest:1.5.4.RELEASE'
   compile 'org.springframework.boot:spring-boot-starter-validation:1.5.4.RELEASE'
+  compile files(org.gradle.internal.jvm.Jvm.current().toolsJar)
 }
 
 compileJava.options.encoding = 'UTF-8'


javatlacati:
Actually if you have a JDK you have already the tools.jar library and I assume you have one since you can compile source code.
Since it's generally a popietary library you need that line to reference it.
How can I test that it works with your setup? Creating a 'JVM runner clone' Kata?

kazk:
> How can I test that it works with your setup?

I would write a test in `test/runners/java_spec.js` similar to spring or add some in `examples/java.yml` showing your ideas.

javatlacati:
Wel actually I can't include Spring or any library due to [lack of knowledge](https://github.com/Codewars/codewars-runner-cli/issues/547).

javatlacati:
Your current setup can already compile so I'm clossing this since it's unnecessary.

PR #490, opened 2017-08-07
Add Kotlin/Groovy/Scala support

kazk:
### Kotlin 1.1.4

Test Frameworks
- JUnit 4 <http://junit.org/junit4/>
- KotlinTest <https://github.com/kotlintest/kotlintest>

Assertion Libraries
- kotlin.test <https://kotlinlang.org/api/latest/kotlin.test/kotlin.test>
- Kluent <https://github.com/MarkusAmshove/Kluent>
- Expekt <https://github.com/winterbe/expekt>

### Groovy 2.4.12

Test Frameworks
- JUnit 4 <http://junit.org/junit4/>
- Spock <http://spockframework.org>

### Scala 2.12.3

Test Frameworks
- JUnit 4 <http://junit.org/junit4/>
- ScalaTest <http://www.scalatest.org/>

---

- [x] Examples
- [x] Tests
- [x] Documentations
- [x] Update README

---

Closes #9 
Closes #165
Closes #175

PR #494, opened 2017-08-11
[POC] Java in gradle-runner

kazk:
This shows how Java can live in `gradle-runner`.
Created `lib/runners/java2.js` on top of #490 (ignore the commits except for the last).

- No hack for `run`/`test`
- Split files support
- Spring Boot support
- Should be compatible

Once we have full `opts.files` support, we should be able to do something like writing tests for Java web apps using Spock (Groovy) as well.


kazk:
> If the runtime doesn't change

This one doesn't have `build.gradle` manipulation because I couldn't understand it well (both the code and the purpose). Also, some dependencies are omitted for simplicity and because of duplication.

OverZealous:
Just to be clear, I accidentally wrote _runtime_ but I meant _run time_, as in how long it takes to run. 😁

jhoffner:
This looks really good!

> This one doesn't have build.gradle manipulation because I couldn't understand it well (both the code and the purpose).

The idea is that all dependencies that we may want to support are added to the build file, so that when the image is built, the dependences would be loaded onto the image. In this way we use the build file the way we would a Gemfile or package.json. However building with all dependencies, including those that are not requested, leads to longer build times - so I was removing any dependencies that are not necessary.

How is the build/run time on this compared to the original?

kazk:
> However building with all dependencies, including those that are not requested, leads to longer build times

I think this was caused by not specifying the `--offline` flag. This flag specifies that the build should be done without accessing any network resources. Without this, I think Gradle makes some request to online repository for each dependency. If any of the dependencies have dynamic version, e.g. "1.0.+", it might also download the latest available during build.
In `gradle-runner`, the dependencies are downloaded at first and every other call has `--offline` to use the cached dependencies. I don't think there's performance penalties by always using the full `build.gradle`.

> How is the build/run time on this compared to the original?

Looking at the time in Travis build logs, this one is slower. But `java-runner` doesn't set the memory limit and I think that's contributing to the speed difference rather than `build.gradle`. Travis build environment has 7.5 GB of memory so the maximum heap size of JVM is probably set to ~1.9GB (7.5GB/4). `gradle-runner` sets the maximum heap size to 256MB.

kazk:
I added `--project-cache-dir /runner/frameworks/gradle` and now this is as fast as the original.

PR #497, opened 2017-08-11
Remove R installation from alt-runner

kazk:
This was part of https://github.com/Codewars/codewars-runner-cli/pull/469/commits/fc97d4767a8365de645a78d268676a7c590455c9, but apparently got lost while merging.

Maybe we should rename `alt-runner` to `php-runner` and `jvm-runner` to `clojure-runner`.


jhoffner:
> Maybe we should rename alt-runner to php-runner and jvm-runner to clojure-runner.

Before doing this we will need to update the server to be able to check for both variants. Otherwise there can be a gap during deployments where the things get out of sync. Happened yesterday with elixir moving over to the new elixir image.

PR #498, opened 2017-08-11
Add typings/ back

kazk:
Support reference paths under `'/runner/typings/main/ambient/'`


kazk:
I jst noticed the whitespace changes my editor caused :(
For the diff without them, see
https://github.com/Codewars/codewars-runner-cli/pull/498/files?w=1

PR #506, opened 2017-08-18
[Swift] Fix invalid reference `test.framework`

kazk:
Minor bug fix. I wrote `test.framework` instead of `opts.testFramework` for some reason :/
I thought ESLint catches these bugs.


kazk:
I'll add this in #507

PR #507, opened 2017-08-21
Cleanup and Refactor

kazk:
Cleanup and refactoring to improve maintainability. I'd recommend looking at each commit when reviewing.

The biggest change is how runners are defined. A runner is now a simple JavaScript object defining "strategies" and no more `shovel`. The code should be easier to follow because the scope is much clearer. Also this makes it easy to switch to `async/await` in the future. There shouldn't be any change in functionality.


lib/runners/ruby.js
```diff
@@ -81,29 +86,37 @@ function prepareEntryFile(opts, require) {
   // if there is no require and an entryPath is provided than just use that file directly
   if (!require && opts.entryPath) return opts.entryPath;
 
-  var entry = [
-    "`rm -rf /workspace/entry.rb`",
+  const entry = [
+    "`rm -rf /workspace/.entry.rb`",
```
kazk: Fixed the file name missing leading dot.

lib/runners/ruby.js
```diff
@@ -81,29 +86,37 @@ function prepareEntryFile(opts, require) {
...
   }
   else {
+    // TODO is opts.dir necessary? it's hard-coded as /workspace here
     if (opts.setup) {
-      entry.push(`require "${util.codeWriteSync('ruby', opts.setup, opts.dir, 'setup.rb')}"`);
+      const setupFile = path.join(opts.dir, 'setup.rb');
+      fs.outputFileSync(setupFile, opts.setup);
+      entry.push(`require "${setupFile}"`);
       // have the file remove itself from the file system after it is loaded, so that it cannot be read by users trying to solve
       entry.push("`rm -rf /workspace/setup.rb`");
```
kazk: These won't work if `opts.dir != /workspace`.

lib/shovel.js
```diff
@@ -160,7 +168,8 @@ function downloadFromGithub(opts, resolve) {
       bash: `
                 cd ${opts.dir}
                 wget -qO- ${repo} | tar xvz --strip-components=1
-            `
+            `,
+      dir: opts.dir,
```
kazk: `util.codeWriteSync` worked without this because it defaults to `/home/codewarrior` when directory is falsy (`undefined`).

PR #512, opened 2017-08-25
Add Solidity runner

### LINKED ISSUE #433: Solidity/Ethereum Support
https://solidity.readthedocs.io/en/develop/

kazk:
Add Solidity v0.4.13 using Truffle v3.4.8.

Packages

- web3 v0.20.x

Note that this image is independent (not using base-runner). If I'm understanding correctly, those services installed in the base runner can't be used anyways. Please let me know if I'm mistaking.

Closes #433


jhoffner:
This looks really good. Why not reuse the node.docker image though? My main reason for asking is that we may want to provide project support at some point, where you can write a node server to interact with a contract. Something like this: https://medium.com/@codetractio/try-out-ethereum-using-only-nodejs-and-npm-eabaaaf97c80

If you have specific compatibility reasons then a separate docker file works for now though, as we  still need to figure out how we are going to be organizing these containers for custom images later (I'm curious how far we can go with multi-stage builds)

kazk:
> Why not reuse the node.docker image though?
>
> If you have specific compatibility reasons

Yeah, this one requires recent versions of Node and `node-runner` is already having issues with multiple versions of Node using modules installed for one version. It's not broken (yet) because modules are installed for Node 6 and most of them works with Node 8. The other way won't work and this one requires that.

> we may want to provide project support at some point, where you can write a node server to interact with a contract.

The problem is that I don't know anything about Solidty/Ethereum. I can help someone adding though.

PR #513, opened 2017-08-27
Additional Node/Ruby packages

jhoffner:
- added ethereum and enzyme packages to node
- added eventide to ruby


docker/ruby.docker
```diff
@@ -15,7 +15,7 @@ RUN mkdir -p /usr/local/etc \
 	} >> /usr/local/etc/gemrc
 
 ENV RUBY_MAJOR 2.3
-ENV RUBY_VERSION 2.3.0
+ENV RUBY_VERSION 2.3.3
 ENV RUBY_DOWNLOAD_SHA256 ba5ba60e5f1aa21b4ef8e9bf35b9ddb57286cb546aac4b5a28c71f459467e507
```
kazk: You need to change `RUBY_DOWNLOAD_SHA256` too

kazk:
I think we should split `package.json`. Like how I did for Solidity.
Mixing CLI dependencies with packages meant to be used by users is not a good idea. I think there's some conflicts already and many warnings.
Ideally we should have `/workspace/node/$version` for each supported versions, but we can at least start from `/workspace/node`.

kazk:
How do you use these? Can you write simple test cases? I really think we need to clean up packages.
On Codewars, it's already 700MB. We have some packages that don't even work with our setup.

jhoffner:
> Can you write simple test cases?

Yeah I was just added some for the ethereum ones. The enzyme ones could use tests too. 

> How do you use these?

Enzyme was requested by a client. I also want to work on adding an inline packages list to the runner output area (inline documentation), so that its more obvious to content creators what packages they can use - to encourage their usage.

jhoffner:
> I think we should split package.json. Like how I did for Solidity.

Let's do this in another PR, it makes sense from a "its weird to repurpose the package.json file" point of view, though not clear on how using a separate package.json will clear up any warnings we are getting.

kazk:
>  though not clear on how using a separate package.json will clear up any warnings we are getting.

- same package with different versions in `dependencies` and `devDevendencies`
- update CLI dependencies without breaking existing content

kazk:
We also need to lock the versions more strictly.

jhoffner:
I'm closing for now. We can reset this branch and I will submit when my additional changes are ready

PR #515, opened 2017-08-29
Replace listen.js

kazk:
Replace `listen.js` with `entrypoint.sh` to reduce memory usage.

`docker run -d runner` or `docker run -d runner listen` keeps the container running by executing `tail -f /dev/null`. If `/runner/prewarm.sh` is readable, `bash /runner/prewarm.sh` is executed first.

If the command is given and the command is either `run` or `run-json`, `node` is prepended so that `node run`/`node run-json` is executed. This keeps the backward compatibility.


entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -e
+
+# if CMD was empty or `listen`, do `tail -f /dev/null` to keep the container running
+if [ "$#" -eq 0 -o "$1" = 'listen' ]; then
+  set -- tail -f /dev/null
```
jhoffner: what exactly is set doing here? I'm not very familiar with its usage.

jhoffner:
Really clean implementation. Great job keeping it backwards compatible, also really great how we can easily change the command away from `node` without having to override the default `entrypoint`

entrypoint.sh
kazk: `set` sets positional parameters when not setting options. `--` after `set` tells that anything after `--` is not option.

Basically, this file does something like
```ruby
(ARGV.clear; ARGV.push('tail', '-f', '/dev/null')) if ARGV.empty? || ARGV[0] == 'listen'
ARGV.unshift('node') if ARGV[0] == 'run' || ARGV[0] == 'run-json'
```
to manipulate `ARGV`.

PR #516, opened 2017-08-30
Ethereum JS

jhoffner:
This adds a number of libraries to the node.docker image to work with Ethereum. This is different than the `solidity.docker` file, which is optimized for working specifically with testing large custom contracts. For this case, we are working with either inline compiled or example contracts and more focused on using the Ethereum API's. 

## Reservations
These API's seem to have stabilized, but I'm concerned that they may still be in flux still and we will have version issues if too much content is built on top of these. I found a number of outdated examples online. With that said, web3 is now at v1.0 (beta) and is the main way of interacting with the API. 

## Riders
- removed some fringe node modules that are not known to have ever been used, in order to trim things down
- added enzyme modules (adding test for these would be ideal, @Dan-Nolan maybe you can work on that when you update the react packages?)


docker/node.docker
```diff
@@ -144,8 +153,8 @@ ENV USER=codewarrior HOME=/home/codewarrior
 ENV NODE_PATH=/usr/lib/node_modules:/runner/node_modules
 
 # Run the test suite to make sure this thing works
-RUN mocha -t 10000 --recursive test/runners/javascript/
-RUN mocha -t 10000 test/runners/coffeescript_spec.js
-RUN mocha -t 10000 test/runners/typescript_spec.js
+#RUN mocha -t 10000 --recursive test/runners/javascript/
+#RUN mocha -t 10000 test/runners/coffeescript_spec.js
+#RUN mocha -t 10000 test/runners/typescript_spec.js
```
kazk: Why are the tests commented out?

test/runners/javascript/modules/ethereum.js
```diff
@@ -0,0 +1,241 @@
+"use strict";
+
+const expect = require('chai').expect;
+
+const runner = require('../../../runner');
+
+describe('ethereum', function() {
```
kazk: I'm glad I had split `javascript_spec.js`. Tests are much easier to find and write :)

jhoffner: Yeah great idea! One thing I did notice is that the new files no longer have _spec as a suffix, and the naming scheme switched to kebob case. Was that on purpose?

docker/node.docker
jhoffner: Whoops, I meant to add those back in. I knew I was going to forget that.

jhoffner:
What the hell, Travis stopped caring about this PR 😟

jhoffner:
@Dan-Nolan does Enzyme really give users anything they didn't have already? I'm now debating if we should actually accommodate their request.

test/runners/javascript/modules/ethereum.js
kazk: `_spec` seems redundant to me because these are under `test/`, but if you want them back I don't mind. kebab case file names are more common in JS projects and matches rest of the codebase.

kazk:
Apparently angularjs.org is not responding :(

OverZealous:
Looks like they got it back up again (that was annoying this morning—no docs!)

OverZealous:
I restarted the build.

test/runners/javascript/modules/ethereum.js
jhoffner: Ah, I can't keep them straight. I could have swore this repo was underscore cased, but apparently its just the specs. 

The useful thing about `_specs` is that when using file find its nice to see that a file is a spec (in case there are other files named the same), but also because there may be helper files within the directory that are not actually specs. Not a big deal though.

PR #517, opened 2017-08-31
Solidity: Added zeppelin support, a critical library to building contracts

jhoffner:


docker/alt.docker
```diff
@@ -34,8 +34,11 @@ COPY frameworks/php frameworks/php
...
 
+COPY entrypoint.sh entrypoint.sh
+RUN chmod +x entrypoint.sh
+
 USER codewarrior
 ENV USER=codewarrior HOME=/home/codewarrior
 RUN mocha -t 3000 test/runners/php_spec.js
 
-ENTRYPOINT ["node"]
+ENTRYPOINT ["./entrypoint.sh"]
```
kazk: Why does this have `entrypoint.sh` changes?

jhoffner: I'm not sure what happened. I think it was at this point that I starting using origin vs upstream. A rebase might help but I don't want to risk spending the next 30 minutes fixing something if it messes up the history even more.

PR #522, opened 2017-09-01
Gradle Optimization Improvements

jhoffner:
- when-prewarmed utility added to ensure we wait until the starting daemon is ready
- cleaned up test cases
- improved test case prewarm integration
- increased jvm memory

### DIFF
--- a/lib/utils/when-prewarmed.js
+++ b/lib/utils/when-prewarmed.js
@@ -49,5 +49,5 @@ module.exports.clean = clean;
 // sets up mocha, also ensures that done never gets called with an argument
 module.exports.setupMocha = function() {
   before(done => ensure(_ => done()));
-  after(done => clean(_ => done()));
+  after(clean);
 };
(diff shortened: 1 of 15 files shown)


kazk:
`kotlin_spec.js` is terminating early because `/runner/frameworks/gradle/gradle.properties` is gone.

```
Error: ENOENT: no such file or directory, open '/runner/frameworks/gradle/gradle.properties'Unhandled rejection AssertionError: expected '' to equal '42\n'
```
I don't know why the exit code is 0. Maybe caught by mocha and mocha's exit code is the number of failed tests, so error before failure is 0.

kazk:
Groovy/Scala performance is about the same on Travis environment :(

And I think other tests are terminated early too, it's missing `X passing (time)` output.
image

Before
image

Probably related to new `setupMocha`.

lib/utils/when-prewarmed.js
```diff
@@ -0,0 +1,53 @@
...
+}
+
+module.exports = whenPrewarmed;
+module.exports.ensure = ensure;
+module.exports.clean = clean;
+
+// sets up mocha, also ensures that done never gets called with an argument
+module.exports.setupMocha = function() {
+  before(done => ensure(_ => done()));
+  after(done => clean(_ => done()));
```
kazk: `clean` doesn't call `_ => done()`

PR #476, opened 2017-07-26
[TypeScript] Make targets configurable

kazk:
Make `tsc` `--target` configurable through `opts.languageVersion`, e.g., `2.4/ES6`, for simple execution and testing with mocha.
I decided not to touch Karma's configuration for now since it's set to ES5 explicitly.

Closes #418
Closes #380

PR #525, opened 2017-09-05
Fix to enable running TS/Karma with a version

OverZealous:
- Now the `languageVersion` is not passed into the Karma/Node runner code, so we don't get Node version errors.
- Updated Node docs to mention Karma testing on TS

### DIFF
--- a/lib/runners/typescript.js
+++ b/lib/runners/typescript.js
@@ -156,7 +156,11 @@ function prepareKarma(opts, interfaceType, runCode, fail) {
     },
   };
 
-  runKarma(opts, runCode, fail, interfaceType, config);
+  const nodeOpts = Object.assign({}, opts);
+  // run on the default node version
+  delete nodeOpts.languageVersion;
+
+  runKarma(nodeOpts, runCode, fail, interfaceType, config);
 }
 
 function _tsc(opts) {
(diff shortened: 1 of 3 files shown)


lib/runners/typescript.js
```diff
@@ -156,7 +156,11 @@ function prepareKarma(opts, interfaceType, runCode, fail) {
     },
   };
 
-  runKarma(opts, runCode, fail, interfaceType, config);
+  const nodeOpts = Object.assign({}, opts);
+  // run on the default node version
+  delete nodeOpts.languageVersion;
```
kazk: Good catch. I was curious how long ago Node 2.4 was and apparently it doesn't even exist :D (it was io.js 2.4 released 2015-07-17).

If specifying the target for Karma is useful, we can do that too. I didn't touch Karma's configuration in https://github.com/Codewars/codewars-runner-cli/pull/476 because it was explicitly set to ES5.
https://github.com/Codewars/codewars-runner-cli/blob/aeb1d5d7628000217dc79b8a3d0594f9668d4f5d/lib/runners/typescript.js#L142-L147

OverZealous: Yeah, I definitely thought about it. I couldn't imagine a good scenario for setting the Karma ES target. Nobody building websites today can safely target anything newer than ES5, and ES3 is too restrictive for most modern JS.

I'll leave it for something we can add in later if it become useful. As it is, we only support a single version of TS, and I'm not 100% sure how significantly the different target versions change the output in this case.

PR #526, opened 2017-09-19
Various refactoring changes to JavaScript test framework

Souzooka:
Changes include:
* Enforce quotation style
* Replace `var` with `let` or `const`
* Use default arguments where applicable
* Remove "hanging elses"
* Change time arithmetic so it uses `getMilliseconds`
* Collapse a few one-liner functions to arrow functions
* Various other things I probably forgot

Since I changed a lot of things and the script was doing some weird stuff (like calling `Test` inside the `Test` declaration, the script should be tested before being implemented in case there's some weird context thing or something along those lines).

### DIFF
--- a/frameworks/javascript/cw-2.js
+++ b/frameworks/javascript/cw-2.js
@@ -1,26 +1,25 @@
-
 try {
-  var util = require('util'),
-      deepEquals = require('lodash').isEqual,
+  const util = require("util"),
+      deepEquals = require("lodash").isEqual,
       Promise = require("bluebird");
 
-  require('./chai-display');
+  require("./chai-display");
 
-  var fnToString = Function.toString;
+  const fnToString = Function.toString;
   Function.prototype.toString = function() {
     switch (this) {
       case Test.expect:
       case Test.randomNumber:
       case Test.randomize:
       case Test.randomToken:
-        return '[Codewars Code]';
+        return "[Codewars Code]";
 
       default:
         return fnToString.call(this);
     }
   };
 
-  var describing = [],
+  let describing = [],
       async = false,
       asyncIts = null,
       asyncDone = null,
@@ -29,14 +28,14 @@ try {
       afterCallbacks = [],
       alwaysExplain = false;
 
-  process.on('uncaughtException', function(err) {
+  process.on("uncaughtException", function(err) {
     if (async) {
       Test.handleError(err);
       if (asyncDone) asyncDone();
     }
   });
 
-  var describeNext = function() {
+  const describeNext = function() {
     if (asyncIts.length > 0) {
       asyncIts.shift()();
     }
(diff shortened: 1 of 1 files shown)


frameworks/javascript/cw-2.js
```diff
@@ -255,196 +239,160 @@ try {
...
-    fail: function(message) {
-      _expect(false, message);
-    },
-    expect: function(passed, message, options) {
-      _expect(passed, message, options);
+        .replace("at Object.Test.handleError", "");
     },
+    pass: () => _expect(true),
+    fail: message => _expect(false, message),
+    expect: (passed, failMsg, options) => _expect(passed, failMsg, options);
```
kazk: `;` is causing parse error.

frameworks/javascript/cw-2.js
```diff
@@ -45,59 +44,52 @@ try {
...
 
-      if (options.passed) {
+      if (options.hasOwnProperty("passed") && options.passed) {
         options.passed(successMsg, options);
       }
-    }
-    else {
-      failMsg = failMsg || 'Value is not what was expected';
-
+    } else {
```
kazk: > Remove "hanging elses"

```javascript
}
else {
```

This style is preferred by @jhoffner and we have a ESLint rule so please don't change this.

frameworks/javascript/cw-2.js
```diff
@@ -108,9 +100,7 @@ try {
     },
 
     // backwards compatibility
-    format: function(obj, options) {
-      Test.display.format(obj, options);
-    },
+    format: this.display.format,
```
kazk: `this != Test`

frameworks/javascript/cw-2.js
```diff
@@ -121,25 +111,23 @@ try {
       }
       return util.inspect(obj);
     },
-    log: function(msg, opts) {
-      Test.display.log(msg, opts);
-    },
+    log: (msg, opts) => this.display.log(msg, opts),
```
kazk: `this != Test`

frameworks/javascript/cw-2.js
```diff
@@ -255,196 +239,160 @@ try {
...
+    expect: (passed, failMsg, options) => _expect(passed, failMsg, options);
     assertSimilar: function(actual, expected, msg, options) {
       this.assertEquals(Test.inspect(actual), Test.inspect(expected), msg, options);
     },
     assertNotSimilar: function(actual, expected, msg, options) {
       this.assertNotEquals(Test.inspect(actual), Test.inspect(expected), msg, options);
     },
-    assertEquals: function(actual, expected, msg, options) {
-      if (typeof(msg) == 'object') {
+    assertEquals: function(actual, expected, msg, options = {successMsg: "Value == " + Test.inspect(expected)}) {
```
kazk: We can't have default parameters because we still have to support old NodeJS versions.

Souzooka:
Hi Kazk, I've gone over the file again and reverted some changes and made a few other tweaks:

The big changes are as follows:
* Enforce consistent quotation style (any single quotes became double quotes)
* More explicit typechecking and default arguments at top of function body (if applicable for the particular function)
* Explicitly use `getMilliseconds` for time arithmetic
* `Test.expect` is now an alias for `_expect` (was calling `_expect` with same parameters as was passed to it)
* Various other little tweaks which hopefully should make things a little more strict/readable

frameworks/javascript/cw-2.js
```diff
@@ -108,9 +119,7 @@ try {
     },
 
     // backwards compatibility
-    format: function(obj, options) {
-      Test.display.format(obj, options);
-    },
+    format: Test.display.format,
```
kazk: This doesn't work because `Test.display` is still `undefined` at this point.
Please test before opening PR and pushing changes.

frameworks/javascript/cw-2.js
```diff
@@ -46,32 +45,36 @@ try {
...
-    if (passed) {
+    if (typeof options !== "object") {
+      options = {};
+    }
+    
+    if (typeof failMsg !== "string") {
+      failMsg = "Value is not what was expected";
+    }
+    
+    if (passed === true) {
```
kazk: This changes the behavior of `Test.expect`

Souzooka:
Alright, that should be good hopefully, apologies for not checking the ESLint settings.

Unfortunately due to legacy compatibility I wasn't able to make as many changes as I wanted. If CW doesn't need compatibility pre-ES5, perhaps it should be considered making cw-2.js a Babel file and having it transpiled to ES5?

kazk:
@Souzooka
For strictly checking types, I *personally* prefer them but this can be breaking change that I'm not comfortable approving. Especially with this commit being extremely hard to review for me. Changes in quotes made other significant changes hard to see and forced me to scan the file multiple times. Also, rest of the codebase doesn't enforce this (we explicitly avoided this ESLint rule for reasons stated above) and you've missed few anyways.

> If CW doesn't need compatibility pre-ES5, perhaps it should be considered making cw-2.js a Babel file and having it transpiled to ES5?

I think we should consider dropping support since the status has been ["End-Of-Life"](https://github.com/nodejs/Release#release-schedule1) for almost a year now, but there're exiting contents that still uses them.
After looking at this file again, I think we should keep this frozen unless significant bugs are found.

I'd like to close this unless @jhoffner can review this and approves. I don't think I can spend any more time.

Souzooka:
After tinkering through the file for a while, I do agree that it would be better to leave it as is, as it's hard to tell if the littlest change could break compatibility with an existing Kata.

This is doubly true if the file is intended to be dropped in the near future anyways. The other solution, as I've noted, is to possibly rewrite the file in Babel and just transpile it down to ES5.

Having said all that I would agree with closing this pull request right now.

Issue #291, opened 2017-01-04
New Elixir Version


allonhadaya:
Hey, I was recently working on a [kata](https://www.codewars.com/kata/exclamation-marks-series-number-3-remove-all-exclamation-marks-from-sentence-except-at-the-end) and came up with this solution:

```elixir
defmodule Solution do
  def remove(s) do
    # assume s does not contain the unit separator character
    unit_separator = "\x1F"
    s
    |> String.replace_trailing("!", unit_separator)
    |> String.replace("!", "")
    |> String.replace_trailing(unit_separator, "!")
  end
end
```

Turns out that `String.replace_trailing` had a bug fix delivered in [1.2.5](https://github.com/elixir-lang/elixir/releases/tag/v1.2.5).

I'm not sure if this alone is enough of a reason to offer a new version of the language, or how exactly you decide that, but I am under the impression that the language / standard library have made quite a bit of progress this past year [1.2.4 -> 1.3.4](https://github.com/elixir-lang/elixir/compare/v1.2.4...v1.3.4).

Let me know what you think.

ice1000:
The newest is 1.5 https://preview.codewars.com/topics/feature-requests#5997de3dc2bd3a773700007b

DCrow:
Would be great, since newest Elixir also includes new Erlang features

kazk:
If the upgrade doesn't break existing content, then we can just swap the installed version, but this is unlikely.
- Elixir v1.5.0 is 100% compatible with v1.2.4
- ExUnit is also compatible
- Existing content doesn't use deprecated features

I've proposed having versioned images, e.g., having images `elixir-runner:1.2.4` and `elixir-runner:1.5.0`, but this requires some rework and not happening soon.

More quicker solution will be to install both versions in the same image using [asdf](https://github.com/asdf-vm/asdf) with [asdf-elixir](https://github.com/asdf-vm/asdf-elixir), but I haven't looked into them yet.

koskoci:
+1

kazk:
This is on our roadmap. The new system we've been working avoids the issue of breaking existing contents. See https://github.com/Codewars/codewars-runner-cli/pull/528#issuecomment-348716997

kazk:
We'll have Elixir 1.6 available on the new system.
See #566 for future updates/discussions.

PR #528, opened 2017-09-27
update elixir version to 1.5.1

jademcgough:
I tried to use `Enum.chunk_every/2` on a kata, only to realize that the version of Elixir that Codewars is running is nearly two years old.

The language is young and is changing fairly rapidly, so it would be great to update to Elixir 1.5.


kazk:
See https://github.com/Codewars/codewars-runner-cli/issues/291#issuecomment-324186809

jademcgough:
@kazk ah, thanks, guess I could have searched open issues more thoroughly.

Would my PR be sufficient to update the build version? If not, what would you want to see done? I'd be happy to contribute more to get this fixed.

kazk:
Thanks for contributing.
Currently, the only way to add newer version is to install multiple versions and switch them depending on the given option. This is because we can only have single `elixir-runner` image and we need to keep the old versions for existing content. I mentioned [asdf](https://github.com/asdf-vm/asdf) as a possible workaround in the comment for this reason.

The current system has been painful to maintain and doesn't scale well, so we're designing a new one to eliminate this issue in general. We can't give an ETA yet.

kazk:
Closing this because of the reasons I commented above.

The new system that supports multiple versions is coming. Most of the basic features are complete, but we're still working on it. I'm the only one working on it at the moment so it took a month or so to add 36 languages.

Adding a new language version will become as easy as creating a new Docker container image (e.g., `codewars/elixir:1.5`) and adjusting the way to execute if there's any incompatibility. Adjustments can be made without affecting other versions.

For Elixir, I rewrote to use `mix test` so we can enhance it in the future like adding some packages and support multiple files.

PR #530, opened 2017-10-05
Adds core-js to Karma JS tests

OverZealous:


OverZealous:
@jhoffner This should be ready to go, it fixes a bug in converting certain ES6 constructs using Babel. (If you reference certain methods or use something like `for(let item of arr)`, you get cryptic errors).

PR #549, opened 2017-11-15
Python Unittest Unhandled Exception

Dan-Nolan:
In unittest exceptions we'll show the exception class rather than the hardcoded exception string so  `<ERROR::>KeyError: 10` instead of `<ERROR::>Unhandled Exception: 10`

### DIFF
--- a/test/runners/python_spec.js
+++ b/test/runners/python_spec.js
@@ -1,7 +1,20 @@
 var expect = require('chai').expect;
 var runner = require('../runner');
+const exec = require('child_process').exec;
 
 describe('python runner', function() {
+  afterEach(function cleanup(done) {
+    exec([
+      'rm -rf',
+      '/home/codewarrior/*.py',
+      '/home/codewarrior/__pycache__',
+      '/home/codewarrior/test',
+    ].join(' '), function(err) {
+      if (err) return done(err);
+      return done();
+    });
+  });
+
   // These specs are compatible with both Python 2 and 3
   ['2', '3', '3.6'].forEach(lv => {
     describe('.run', function() {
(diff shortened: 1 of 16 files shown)


test/runners/python_spec.js
```diff
@@ -133,6 +133,28 @@ describe('python runner', function() {
           done();
         });
       });
+      it('should show KeyErrors', function(done) {
+        runner.run({
+          language: 'python',
+          languageVersion: lv,
+          testFramework: 'unittest',
+          code: 'a = {}; a[10] += 3',
```
kazk: @Dan-Nolan these `testFramework` and `code` needs to be removed because ESLint is failing with "Duplicate key".

frameworks/python3/unittestwrapper.py
```diff
@@ -27,9 +30,9 @@ def addSuccess(self, test):
     super(CwTestResult, self).addSuccess(test)
 
   def addFailure(self, test, err):
-    print("<FAILED::>" + formatMessage(err[1]))
+    print("\n<FAILED::>" + formatError(err))
     super(CwTestResult, self).addFailure(test, err)
 
   def addError(self, test, err):
-    print("<ERROR::>Unhandled Exception: " + formatMessage(err[1]))
+    print("\n<ERROR::>" + formatError(err))
```
kazk: I don't think this file (or any file in frameworks/python3) is used?

https://github.com/Codewars/codewars-runner-cli/blob/dbf943d62c8d6ed451d27ac1817c0f13b1de558c/lib/runners/python.js#L110

No need to revert the change though.

Dan-Nolan: Yeah I didn't think so either, fixed it just in case it gets used again 😄

test/runners/python_spec.js
Dan-Nolan: Ah thanks! I'll fix this up.

jhoffner:
@Dan-Nolan I guess I broke something when I try to merge the conflict.

kazk:
This is failing because unittest with Python 3 now uses the no-concatenation introduced in #552, so the error output looks like the following:

```
<ERROR::>Unhandled Exception
<LOG:ESC:Error>KeyError: 10
<LOG:ESC:Traceback>  File "/home/codewarrior/test/test_solution.py", line 5, in test_assert<:LF:>    self.assertEqual(go(1), 1)<:LF:>  File "/home/codewarrior/solution.py", line 3, in go<:LF:>    a[10] += 3
```

and doesn't contain `<ERROR::>KeyError: 10`

kazk:
Closing because this is no longer used.

PR #552, opened 2017-12-06
Add Python 3 unittest runner without concatenation

kazk:
This is based on #347, but tries to be drop-in replacement. It passes all the tests, but needs further tests against existing contents.
Only Python 3 for now. I'm hoping this is compatible with Python 2 without much change.
I remember having some issue when doing #347, but I don't remember the details since it's been a while and I didn't document them :|

---

Features:

- No file concatenation
  - Prepends few lines if necessary to support existing contents. See below.
- Show file name and line numbers
- Test duration for each block
- Arbitrary test class name
- Multiple test suites
  - Needs more work.  Tests are loaded, but combined into one suite

---

To be backwards compatible, few lines are prepended when necessary:

- prepend `import unittest` to fixture if missing
- prepend `from solution import *` to fixture to simulate concatenation
- prepend `from setup import *` to solution to simulate concatenation

Line numbers for `solution.py` can be off by at most one. We can add additional message to warn candidates, but improved error message with stack trace makes the error location obvious.
The stack trace contains something like this:
```
  File "/home/codewarrior/solution.py", line 1
    def add(a, b): return a b
                            ^
SyntaxError: invalid syntax
```

Line numbers for fixture can be off by at most two, but this is usually not useful for candidates and authors should know why they're off.

### DIFF
--- a/frameworks/python/codewars.py
+++ b/frameworks/python/codewars.py
@@ -1,0 +1,71 @@
+import unittest
+import traceback
+from time import perf_counter
+
+class CodewarsTestRunner(object):
+    def __init__(self): pass
+    def run(self, test):
+        r = CodewarsTestResult()
+        s = perf_counter()
+        print("\n<DESCRIBE::>Tests")
+        try:
+            test(r)
+        finally:
+            pass
+        print("\n<COMPLETEDIN::>{:.4f}".format(1000*(perf_counter() - s)))
+        return r
+
+__unittest = True
+class CodewarsTestResult(unittest.TestResult):
+    def __init__(self):
+        super().__init__()
+        self.start = 0.0
+
+    def startTest(self, test):
+        print("\n<IT::>" + test._testMethodName)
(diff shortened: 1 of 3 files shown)


kazk:
One limitation with this is that identifiers starting with `_` are not exported, so it won't work. I don't think there are many, but this might break some.

frameworks/python/codewars.py
```diff
@@ -0,0 +1,71 @@
...
+        while tb and self._is_relevant_tb_level(tb):
+            tb = tb.tb_next
+        if exctype is test.failureException:
+            length = self._count_relevant_tb_levels(tb) # Skip assert*() traceback levels
+        else:
+            length = None
+        return ''.join(traceback.format_tb(tb, limit=length))
+
+    def _is_relevant_tb_level(self, tb):
+        return '__unittest' in tb.tb_frame.f_globals
```
Dan-Nolan: Good idea 👍

Issue #569, opened 2018-01-29
Refactor and restructure Python test framework


Voileexperiments:
Currently Python tests are run like this:

```python
test.describe('tests')
test.it('should pass')
test.assert_equals(mul(2,3), 6)
test.it('should pass too')
test.assert_equals(add(2,3), 5)
test.describe('more tests')
# ...
```

Which is not ideal because:
- No automatic closing of test blocks, which means we have to `print('<COMPLETEDIN::>')` in order for multiple `describe`/`it` blocks to be nested properly in the output display (and to the time display, currently we have to do the benchmarking ourselves :V)
- Unlike other language versions, there are no closures, we have to manually wrap it inside another function to not leak everything to global scope
- Similarly, no support for block-specific features (e.g `Test.describe` in JS accepts optional override arguments for timeouts, `Test.before`, ...)
- Not consistent with other languages (see point 2)
- Encourage people to disregard proper formatting and indent (seriously, who wants to read flat code? As if we didn't have enough crappy kata makers already...)
- etc etc etc...

To make Python test framework consistent with other languages, properly introduces scope/closure, and more natural to use/more managable, I'd suggest using `with` as kind of a context manager:

```python
with test.describe('tests') as d:
	with test.it('should pass') as f:
		test.assert_equals(mul(2,3), 6)
	with test.it('should pass too') as f:
		test.assert_equals(add(2,3), 5)
with test.describe('more tests') as d:
	# ...
```

Then we can open and close blocks effortlessly via the use of `__enter__` and `__exit__` (and do benchmarking too). They can also provide additionally arguments for some customizations.

(There are lots of things that can be changed, e.g maybe we can utilize those `d` and `f` to add some extra functionality, or refactor some test method to them? I don't know, but at least this seems infinitely better than what Python test framework is structure right now. Currently it's like, well, chaos that could only come from JS.)

kazk:
Another issue is that `cw-2.py` terminates on first failure by default and the default behavior is inconsistent even within itself. It terminates unless the test used `expect` (`allow_raise=False`) :/

https://github.com/Codewars/codewars-runner-cli/blob/dbf943d62c8d6ed451d27ac1817c0f13b1de558c/frameworks/python/cw-2.py#L11
https://github.com/Codewars/codewars-runner-cli/blob/dbf943d62c8d6ed451d27ac1817c0f13b1de558c/frameworks/python/cw-2.py#L21
[...]

Issue #558, opened 2017-12-20
Encoding Issue in Python3 Runner


suic86:
I'm trying to fix an issue in [this kata](https://www.codewars.com/kata/hungarian-cardinal-numerals/python) but I'm getting the following error in Python3: 

```python
Unable to decode the command from the command line:
UnicodeEncodeError: 'utf-8' codec can't encode character '\udcc5' in position 262: surrogates not allowed
```

kazk:
Are you trying to print something? If so, what happens if you use `.encode("utf-8", "surrogateescape")`?

Can you post the string you're trying to work with?

---

I'll look into Python encoding issues. It's been also reported for Python 2 in #508 and #545.
It might be possible to fix by adjusting environment variables.

```python
import locale
import sys
print(locale.getdefaultlocale())
print(sys.stdout.encoding)
```
from https://wiki.python.org/moin/PrintFails

Python 2.7 outputs:
```
(None, None)
None
```

Python 3.4/3.6 outputs:
```
(None, None)
ANSI_X3.4-1968
```

suic86:
Hi,

I've tried `.encode("utf-8", "surrogateescape")` as I thought that it was because of `print()`. However, taking out the `print`s didn't help. I went through the "encoding exercise" about 3 times. The discussions under my Hungarian language katas contains more information.

**I guess that:**

1. **the issue I have now happens when the code is read before execution** so this is IMO an issue with the runner/docker container so I can't fix it in my code.
2. the issues with printing are caused by different environment variables on different machines. It happened to me, that two different runs on the same version of Python printed different results. I know about `locale.getdefaultlocale()` as I tried them at the time I created those katas. Even these gave different result on different runs.

Thanks,

suic

kazk:
@suic86
Your "Hangarian Cardinal Numerals" works with Python 3.4 by changing the following.

```diff
- print "%s -> %s" % (inp, exp.encode('latin-1'))
+ print("{} -> {}".format(inp, exp))
```

```diff
- print "%s -> %s" % (inp, exp.encode('latin-1'))
+ print("{} -> {}".format(inp, exp))
```

---

I'll try fixing how Python files are handled so
```python
# -*- coding: utf-8 -*-
```
will work as expected for Python 2.

kazk:
I've fixed and deployed the changes.

For Python 2 to work with Non-ASCII characters in source, [the magic comment](https://www.python.org/dev/peps/pep-0263/) is required

```
[...]

Issue #545, opened 2017-11-03
Python does not support special characters such as ♠♥♣♦


MarkSchemmer:
Python does not support such characters in the CW console and throws an error, every time I try to compare or use the characters.

MarkSchemmer:
The following characters such as ♠♥♣♦

Here is an link to my translation [here please view](https://www.codewars.com/kumite/59fcbb016975983aa5000149?sel=59fcbb016975983aa5000149)

The error that is thrown is ```Error: python version python not supported```

MarkSchemmer:
When you have the chance please PM me in the Gitter Channel I would like to help fix this issue if possible.

kodaman2:
Just a suggestion why not use letters for the suits? H, S, D, C. Easier than dealing with unicodes across python versions.

MarkSchemmer:
@kodaman2, That is actually a really good idea!

MarkSchemmer:
I'm going to change the Kata and I will ask @myjinxin2015 about it as well!

kazk:
Fixed with https://github.com/Codewars/codewars.com/issues/1336

Python 3:

```python
# any of these will work
print("♠♥♣♦")
print("\u2660\u2665\u2663\u2666")
print(u"♠♥♣♦")
print(u"\u2660\u2665\u2663\u2666")
```

Python 2:
```python
print u"\u2660\u2665\u2663\u2666"
```

Issue #508, opened 2017-08-22
cw-2.py throws UnicodeEncodeError


Voileexperiments:
I'm doing this kata today:

https://www.codewars.com/kata/so-many-kinds-of-numbers/discuss

Turns out that the test methods message reports are using `ascii` as the encode charset, and hence will cause problems when trying to print out unicode characters.

```
Traceback:
   in <module>
  File "./frameworks/python/cw-2.py", line 28, in assert_equals
    expect(actual == expected, message, allow_raise)
  File "./frameworks/python/cw-2.py", line 16, in expect
    print("\n<FAILED::>{0}".format(message))
UnicodeEncodeError: 'ascii' codec can't encode character '\xb2' in position 13: ordinal not in range(128)
```

SpecterTerrasbane:
Coming here from [#459](https://github.com/Codewars/codewars.com/issues/459), regarding the [Vigenere Cipher Helper](https://www.codewars.com/kata/vigenere-cipher-helper) kata:

I don't know if this helps, but running against tests with unicode in the _authoring_ editor works just fine, while running the exact same solution code in the _training_ editor causes the issue that everyone is complaining about in the comments on that kata.  Observed when I attempted to edit the kata to see what was going on with the Katakana test.

kazk:
Fixed in https://github.com/Codewars/codewars.com/issues/1336

```python
test.assert_equals("x", u'\u2713')
```
Python 3.4/3.6
```
'x' should equal '✓'
```
Python 2.7
```
'x' should equal u'\u2713'
```

PR #571, opened 2018-02-10
Improved cw-2.py

Bubbler-4:
This PR introduces many improvements based on the discussions at #569 and [my own Kumite](https://www.codewars.com/kumite/5a724cec4a6b34f5d00000ba?sel=5a724cec4a6b34f5d00000ba).

---

# Testing Functions 

* Make testing functions non-blocking
* Change the expression inside `assert_not_equals` so that it can prevent operator injection hacks
* Provide `pass_`, `fail`, and `assert_approx_equals` with customizable precision (`pass` is a Python keyword)
* Replace hardcoded `print`s with `display`

# Describe and It

* Build the decorator version of `describe` and `it`, so the text fixture may look like this (and the decorator itself runs the code, so it doesn't need a separate runner function):

```python
@describe('describe-text')
def describe1():
    @it('it-text', before=f1, after=f2)
    def it1():
        # some test function
    @it('it-text')
    def it2():
        # some test function
```

* Properly close `describe` and `it` blocks in the test output
* Print the running times of `describe` and `it` blocks
* Provide `before` and `after` for `describe` and `it`

# Timeout Utility

* Apply a timeout limit on any block of code (the parameter is in seconds, and floats are accepted):

```python
@timeout(0.01)
def some_block():
    do_something()
    test_something()
```

* This is guaranteed to generate a failed test when the block timeouts

# The Unicode Error Trap

* Print using `"\uXXXX"` and `"\UXXXXXXXX"` literal style for tagged messages
* Print using `"&#XXXX;"` HTML escape style for simple logs
* This addresses #508 and #545, but not #558 (which, I guess, could be solvable by providing the source charset on the first line, which in turn requires some edits in the runner code, not `cw-2.py`)

### DIFF
--- a/frameworks/python/cw-2.py
+++ b/frameworks/python/cw-2.py
@@ -1,12 +1,12 @@
 from __future__ import print_function
-import six
+import re, six
 
 class AssertException(Exception):
     pass
 
 
 '''Fix the dreaded Unicode Error Trap'''
-def print(*args, **kwargs):
+def uni_print(*args, **kwargs):
     from sys import stdout
     sep = kwargs.get('sep', ' ')
     end = kwargs.get('end', '\n')
@@ -16,9 +16,10 @@ def print(*args, **kwargs):
         if ord(c) >= 128: return u'&#{};'.format(ord(c))
         return c
     def _escape(s):
-        if isinstance(s, six.string_types):
-            return ''.join(_replace(c) for c in s)
-        return s
+        escaped = ''.join(_replace(c) for c in six.text_type(s))
+        escaped = re.sub(r'\\u([\da-f]{4})', lambda m: '&#{};'.format(int(m.group(1), 16)), escaped)
+        escaped = re.sub(r'\\U([\da-f]{8})', lambda m: '&#{};'.format(int(m.group(1), 16)), escaped)
+        return escaped
         
     six.print_(*map(_escape, args), sep=_escape(sep), end=_escape(end), file=file)
 
--- a/frameworks/python/cw-2.py
+++ b/frameworks/python/cw-2.py
@@ -15,8 +15,11 @@ def print(*args, **kwargs):
     def _replace(c):
         if ord(c) >= 128: return u'&#{};'.format(ord(c))
         return c
-    def _escape(s): return ''.join(_replace(c) for c in s)
-    
+    def _escape(s):
+        if isinstance(s, six.string_types):
+            return ''.join(_replace(c) for c in s)
+        return s
+        
     six.print_(*map(_escape, args), sep=_escape(sep), end=_escape(end), file=file)
 
 
--- a/frameworks/python/cw-2.py
+++ b/frameworks/python/cw-2.py
@@ -64,9 +64,54 @@ def assert_approx_equals(actual, expected, margin=1e-9, message=None, allow_rais
     expect(abs((actual - expected) / div) < margin, message, allow_raise)
 
 
-def describe(message):
-    display('DESCRIBE', message)
+'''
+Usage:
+@describe('describe text')
+def describe1():
+    @it('it text')
+    def it1():
+        # some test cases...
+'''
+def _timed_block_factory(opening_text):
+    from timeit import default_timer as timer
+    from traceback import format_exception
+    from sys import exc_info
+    
+    def _timed_block_decorator(s, before=None, after=None):
+        display(opening_text, s)
+        def wrapper(func):
+            if callable(before): before()
+            time = timer()
+            try: func()
+            except:
+                fail('Unexpected exception raised')
+                tb_str = ''.join(format_exception(*exc_info()))
+                display('ERROR', tb_str)
+            display('COMPLETEDIN', '{:.2f}'.format((timer() - time) * 1000))
+            if callable(after): after()
+        return wrapper
+    return _timed_block_decorator
+
+describe = _timed_block_factory('DESCRIBE')
+it = _timed_block_factory('IT')
 
 
-def it(message):
-    display('IT', message)
+'''
+Timeout utility
+Usage:
+@timeout(sec)
+def some_tests():
+    any code block...
+Note: Timeout value can be a float.
+'''
+def timeout(sec):
+    def wrapper(func):
+        from multiprocessing import Process
+        process = Process(target=func)
+        process.start()
+        process.join(sec)
+        if process.is_alive():
+            fail('Exceeded time limit of {:.3f} seconds'.format(sec))
+            process.terminate()
+            process.join()
+    return wrapper
--- a/frameworks/python/cw-2.py
+++ b/frameworks/python/cw-2.py
@@ -8,17 +8,21 @@ def format_message(message):
     return message.replace("\n", "<:LF:>")
 
 
+def display(type, message, label="", mode=""):
+    print("\n<{0}:{1}:{2}>{3}".format(type.upper(), mode.upper(), label, format_message(message)))
+
+
 def expect(passed=None, message=None, allow_raise=False):
     if passed:
-        print("\n<PASSED::>Test Passed")
+        display('PASSED', 'Test Passed')
     else:
         message = message or "Value is not what was expected"
-        print("\n<FAILED::>{0}".format(message))
+        display('FAILED', message)
         if allow_raise:
             raise AssertException(message)
 
 
-def assert_equals(actual, expected, message=None, allow_raise=True):
+def assert_equals(actual, expected, message=None, allow_raise=False):
     equals_msg = "{0} should equal {1}".format(repr(actual), repr(expected))
     if message is None:
         message = equals_msg
@@ -28,15 +32,14 @@ def assert_equals(actual, expected, message=None, allow_raise=True):
     expect(actual == expected, message, allow_raise)
 
 
-def assert_not_equals(actual, expected, message=None, allow_raise=True):
-    equals_msg = \
-        "{0} should not equal {1}".format(repr(actual), repr(expected))
+def assert_not_equals(actual, expected, message=None, allow_raise=False):
+    equals_msg = "{0} should not equal {1}".format(repr(actual), repr(expected))
     if message is None:
         message = equals_msg
     else:
         message += ": " + equals_msg
 
-    expect(actual != expected, message, allow_raise)
+    expect(not (actual == expected), message, allow_raise)
 
 
 def expect_error(message, function):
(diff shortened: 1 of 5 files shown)


Bubbler-4:
@kazk Would it be OK to add runner-side changes upon this, if this PR isn't going to be merged soon?

kazk:
@Bubbler-4 Thanks! Looks good to me so far :)
Can you elaborate on "runner-side" changes? If you mean [the fix to avoid concatenation](https://github.com/Codewars/codewars-runner-cli/issues/569#issuecomment-362130262), then I'm interested.

test/runners/python_spec.js
```diff
@@ -225,3 +225,102 @@ describe('Output format commands', function() {
     });
   }
 });
+
+describe('Fixed and new features', function() {
```
kazk: Can you also add the test for the Unicode fix (#508, #545)?

kazk: > could be solvable by providing the source charset on the first line, which in turn requires some edits in the runner code

Do you mean adding `# coding: utf-8`?

Bubbler-4: Yes. But since we're going for multiple files approach, we'll need it in each file (at least preloaded, solution, and fixture).

kazk: If we go multiple files, can users just add them when needed?
This is currently impossible because we concatenate them and the magic comment is only valid if it's in the first or the second line.

Bubbler-4: Since we'll want to prepend something to the code when needed (just like in your `unittest` fix), I'd just add it to every file I create. At least it won't break existing things (because utf-8 is a superset of ascii).

Bubbler-4:
For the runner changes, I'm thinking about these (blatantly copied from my kumite's comment section):

* Add to the first line `# -*- coding: utf-8 -*`- so we can use Unicode literals without `\u` escape
* Write each source file into `.py` files using `writeFileSync` and then import them in the runner code
* Remove `sanitizeStdErr` since its existence isn't relevant anymore, and giving full stack trace is much better choice
* Probably we can also write `runner.py` and run with `python runner.py` instead of `python -c ...` which is ugly af
* We don't even need to import `cw-2.py` to `runner.py`; what if we do so in `fixture.py`?

So it covers both [the discussion here](https://github.com/Codewars/codewars-runner-cli/issues/569#issuecomment-362130262) and #558, but it'll be really breaking changes.

kazk:
We can't have any breaking changes for this repository. Which of those are breaking changes?

`# -*- coding: utf-8 -*-` and writing files looks fine to me. And removing `sanitizeStdErr` is technically a breaking change, but shouldn't have effects so I think we should be fine.


> - We don't even need to import `cw-2.py` to `runner.py`; what if we do so in `fixture.py`?

Yeah, I dislike how some languages runner do these. Also that `reload` and path manipulation shouldn't be necessary if we set `PYTHON_PATH` appropriately, right?

I think moving them to `fixture.py` is fine. My main concern with concatenation is line numbers in error messages not matching their solution.

kazk:
Python 2.7 is printing `'\uac00\n'` when expected to include `'&#44032;'`
Python 3.4, 3.6 is throwing:
```
UnicodeEncodeError: 'ascii' codec can't encode character '\uac00' in position 0: ordinal not in range(128)
```

test/runners/python_spec.js
```diff
@@ -322,5 +322,31 @@ describe('Fixed and new features', function() {
...
       });
     });
+
+    it(`should support unicode output (log) (Python${v} cw-2)`, function(done) {
+      runner.run({
+        language: 'python',
+        languageVersion: v,
+        testFramework: 'cw-2',
+        code: 'a = 1',
+        fixture: 'print("u\\uac00")',
```
kazk: Did you mean `fixture: 'print(u"\\uac00")',`? `"u` vs `u"`. Not sure if that makes any difference.

frameworks/python/cw-2.py
```diff
@@ -1,24 +1,49 @@
 from __future__ import print_function
+import six
 
 class AssertException(Exception):
     pass
 
 
+'''Fix the dreaded Unicode Error Trap'''
+def print(*args, **kwargs):
```
kazk: This `print` doesn't seem to work when given non-string, e.g., `print(1)` throws
```
TypeError: 'int' object is not iterable
```
in Python 2.7 and
```
TypeError: print() argument after * must be a sequence, not map
```
in Python 3

frameworks/python/cw-2.py
```diff
@@ -1,24 +1,49 @@
...
+    from sys import stdout
+    sep = kwargs.get('sep', ' ')
+    end = kwargs.get('end', '\n')
+    file = kwargs.get('file', stdout)
+    
+    def _replace(c):
+        if ord(c) >= 128: return u'&#{};'.format(ord(c))
+        return c
+    def _escape(s): return ''.join(_replace(c) for c in s)
```
kazk: In Python 2, the above error is from this `_escape`.

test/runners/python_spec.js
kazk: For Python 2.7, fixing this should make the test pass.
```python
print(u"\uac00") #=> &#44032;
print("u\uac00") #=> u\uac00
```

Bubbler-4:
I'm out today, so changes won't be made until tomorrow.

kazk:
Don't worry, take your time. I just wanted to leave some feedback because I don't know if I have the time to review during the week.

kazk:
@Bubbler-4 
For Python 3 throwing error, setting environment variable `PYTHONIOENCODING=UTF-8` might resolve the issue and it might not be necessary to escape them.

- https://stackoverflow.com/a/43989185
- https://chase-seibert.github.io/blog/2014/01/12/python-unicode-console-output.html

Setting environment variable can be done by changing
https://github.com/Codewars/codewars-runner-cli/blob/dbf943d62c8d6ed451d27ac1817c0f13b1de558c/lib/runners/python.js#L37

to

```javascript
    run({
      name: pythonCmd(opts),
      args: ['-c', code],
      options: {
        env: Object.assign({}, process.env, {PYTHONIOENCODING: 'UTF-8'})
      }
    });
```

Bubbler-4:
Finally solved it, without using environment variables:

* Fall back to providing `uni_print` instead of overwriting `print` (I forgot Py2 runner doesn't have `from __future__ import print_function` at the beginning)
* Force cast to `six.text_type`, then apply some regex for better output in Py2

test/runners/python_spec.js
```diff
@@ -329,9 +329,12 @@ describe('Fixed and new features', function() {
         languageVersion: v,
         testFramework: 'cw-2',
         code: 'a = 1',
-        fixture: 'print(1, "2", u"\\uac00")',
+        fixture: 'test.uni_print(1, "a", u"\\uac00", [314159, "b", u"\\uac01"])',
```
kazk: For Python 2, this is printing:
```
1 a &#44032; [314159, 'b', u'&#44033;']
```
image

https://travis-ci.org/Codewars/codewars-runner-cli/jobs/340822289#L1401

Bubbler-4: Should I remove the `u` flag in front of the literal? I think I'll leave it there, because `str` and `unicode` are two different things in Py2.

kazk: Ok, I didn't realize that was intentional. So it'll be rendered like this:
image

Bubbler-4: Yes, I tested the output using my kumite on CW, but I don't know how it'll be displayed on Qualified.

kazk:
Looks good to me! I'll try to do some testing against existing contents during this week.
Great job on keeping it backwards compatible.
I'm little concerned that the `print("<COMPLETEDIN::>")` workaround some users have been doing will confuse the test output parser though. Hopefully there's not too many and power users can help with finding and removing them.

And thanks for looking into the Unicode issue and `uni_print`!

Bubbler-4:
The workaround should work just fine because, if `describe` or `it` is used old-style, it does not emit `<COMPLETEDIN::>` at all:



Glad you like it :smile:

kazk:
@Bubbler-4 Sorry for the delay. I didn't have the time to test this last week.
Just a quick update. I've deployed this to the new runner environment that I've been testing against mirrored CW traffic and haven't found compatibility issues so far :+1:

kazk:
I found the root cause of the Unicode issue: https://bugs.python.org/issue19846
The image we currently use on Codewars has no locale set (`LC_CTYPE="POSIX"`), forcing ASCII :(

The new runner uses [the official Docker image](https://hub.docker.com/_/python/) with (`LANG C.UTF-8`) and we can just print Unicode without issues.
For Python 2 image, I've also set `PYTHONIOENCODING=UTF-8` because `sys.stdout.encoding` was `None` (setting this will do `.encode("utf-8")` implicitly if I'm understanding correctly).
Overriding the encoding of `stdin`/`stdout`/`stderr` might cause issues for daily use, but should be fine for our use case.

Now we have `print(u'\u2713')` in Python 2.7/3.4/3.6 output `✓` :tada:

Voileexperiments:
When will this be employed to the main CW site (or preview site)? I miss new changes :(

kazk:
I haven't found anything major (looking at the number of successful Python completions) so this should be safe to merge.
@jhoffner what do you think?

If we do, try not to use `test.uni_print`. It's no longer necessary on the new runner so I'll remove that eventually. Current timeline to deploy the new runner for CW is by the end of next week.

kazk:
@Bubbler-4 
preview.codewars.com uses the new runner now if you want to try using this on Kumite. Just be careful not to publish a kata with it.
Thanks for your work!

Bubbler-4:
I just made an [example kumite](https://preview.codewars.com/kumite/5aaa2d33f51c8a313b00000e?sel=5aaa2d33f51c8a313b00000e), and the new framework works perfectly :+1:

Just one thing is, it looks like the runner overhead has increased by ~500ms. It shouldn't be too much of a concern except a few katas that are insanely tight on the bound, but maybe you can consider increasing the time limit to 12500ms.

kazk:
Thanks for trying it out. I realized that `describe`s can be nested too :+1:

The issue with printing Unicode should be resolved as well. I've also avoided concatenation so the error message should be more clear. But I might need to revert it because I noticed some kata uses test methods in preloaded section.

---

Yeah, the wall time (time displayed at the top) won't be the same because the time includes some extra steps. The actual test execution time (time inside test results) shouldn't be that different.

I might make some adjustments later if this is a major problem, but overall the performance should be better and more stable for most of the languages. I'm not too worried because the old runner have inconsistent performance and the wall time can vary more than 500ms.

Bubbler-4:
> But I might need to revert it because I noticed some kata uses test methods in preloaded section.

How about prepending `cw-2.py` import to `setup.py` as well? It's not exposed to the solver, so it doesn't matter if its line numbers are correct or not. Also, that part is the single greatest improvement for kata solvers, so we can't give it up so easily.

kazk:
Yeah, concatenation is terrible for UX. I'll prepend `test = Test = __import__('cw-2')` to `setup.py`.

Bubbler-4:
@kazk 
Blind4Basics made a suggestion to improve `expect_error` so that it can target a specific exception class. [Here](https://preview.codewars.com/kumite/5ab735bee7093b17b2000084?sel=5ab735bee7093b17b2000084) is my kumite that implements the idea, along with some test cases. Is it possible to add this to the new runner?

kazk:
I've added it with slight modification. `expect_error(message, function, exception=Exception)`.

I'll deploy it maybe tomorrow along with other languages that I need to fix.

kazk:
@Bubbler-4 `expect_error` is on preview now, see [this fork](https://preview.codewars.com/kumite/5ab840202619c8f251000029?sel=5ab840202619c8f251000029).

kazk:
I'm going to merge this so @Bubbler-4 gets some credit for it. The actual source that was derived from this will be open sourced later.

Blind4Basics:
Hi guys,

I might have found a way to efficiently forbid the import of modules in python (I doubt it will be bullet proof, I'm not enough in the deep structure of python to know. Though, it seems at least not easy to crack). That could be a nice addition to the test framework, I believe, in the following manner:

* function/data stored in the file of the framework
* a tuple of strings available for the user, each string being the name of a module to forbid (some comments in addition to explain the functioning of it)
* a bit of code to execute just after the preloaded section has been imported and before the import of the file of the user.

I'd send to both of you (@Bubbler-4, @kazk) the code by MP on gitter (I'd rather like that it's not visible for now), so that you can tell me if you think it's worth of the addition.

Cheers,
B4B

PR #835, opened 2020-09-10
Turn off term rewriting macros for Nim

metagn:
Term rewriting macros make it easy to cheat Nim tests. There are no katas that require their use and they're an experimental feature in the first place.

### DIFF
--- a/lib/runners/nim.js
+++ b/lib/runners/nim.js
@@ -53,6 +53,7 @@ function nimArgs(file) {
     '--stackTrace:on',  // turn stack tracing on
     '--lineTrace:on',   // turn line tracing on
     '--checks:on',      // turn all runtime checks on
+    '--trmacros:off',   // turn off term rewriting macros
     '--path:/runner/frameworks/nim',
     file
   ];


kazk:
Thanks for trying to contribute. Unfortunately, this repository only exists for the issue tracker and the code is no longer used.

image

---

But I can apply this change to the current runner.
Can you elaborate on the effect of this change? Any example of a cheat this prevents?

g964:
@hlaaftana posted an issue at <https://www.codewars.com/kata/5b1cd19fcd206af728000056/discuss#5f59fdd701e9c60029deae07>
Maybe it can help.

hlaaftana:
It's not for every test, but it's very easy to write a test that's susceptible to tampering with. g964's katas that check floats have a check that basically does `abs(actual - expected) <= epsilon`, but there's nowhere else in the test where `<=` should return false, so you can override all `<=` calls to make all the tests pass like so:

```nim
template replaceLe*{`<=`(a, b)}(a, b: float64): bool = true
```

Something I just found out is that there is a much more effective solution at breaking katas. You can define your own `check` overload and the tests can call it, because the `check` macro in Nim's unittest has an underspecified type for the first argument (`untyped`). You can change it to bool like so and it will work: 

```nim
template check*(a: bool) = discard
```

The solution is either every Nim test calling `unittest.check` every time, or simply changing `import solution` in the test preload to `import solution except test, suite, check`. The change would be in this line:

https://github.com/codewars/codewars-runner-cli/blob/d1c1d8bfb9919ee61d520370484894be27f48739/lib/runners/nim.js#L24

kazk:
Thanks. I tested the changes against all Nim kata and it didn't break any of the reference solutions. I'll let you know when it's deployed.

kazk:
Forgot to update, but this was deployed. Thanks for contributing!

